PR #35, opened 2015-11-24
Cleanup GUIs all around:

tterrag1098:
- Fix lighting errors throughout
- Add item highlights
- Fix buttons being cut off and uncentered
- Add a background to the headers on the recipe GUI
- Change the color of the header text on the recipe GUI


mezz:
Thanks tterrag! merged

PR #47, opened 2015-12-03
Update ru_RU.lang

Adaptivity:
Hope 1.8 branch is correct for PRing.


mezz:
Thanks for the PR!
Either branch is fine right now, I'm juggling both so they match.

PR #60, opened 2015-12-12
Banish the access transformers back to the hell they came from

shadowfacts:


mezz:
Honestly this is a lot just to get around ATs.
What's so evil about the ATs compared to this? :/

asiekierka:
Aren't access transformers faster than reflection?

Vexatos:
@asiekierka A lot faster.

mezz:
Sorry @shadowfacts, I appreciate the effort though.

PR #67, opened 2015-12-15
Extend German translation

wlhlm:
Moar translations! :smile:

This is for 1.8.8. Will you merge them back for 1.8, or should I make a separate PR?

A thing I've noticed from the [English strings](https://github.com/mezz/JustEnoughItems/blob/1.8/src/main/resources/assets/jei/lang/en_US.lang#L7):

```
jei.tooltip.error.recipe.transfer.inventory.full=Inventory is too full
```

What is the difference between being full and too full?

### DIFF
--- a/src/main/resources/assets/jei/lang/de_DE.lang
+++ b/src/main/resources/assets/jei/lang/de_DE.lang
@@ -1,8 +1,62 @@
+# Tooltips
+jei.tooltip.config=Konfiguration
+jei.tooltip.show.all.recipes=Zeige alle Rezepte
+
+# Error Tooltips
+jei.tooltip.error.recipe.transfer.missing=Fehlende Items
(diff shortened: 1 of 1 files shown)


mezz:
Thanks! I will make sure it ends up on both branches. 
1.8 is the main branch right now and I copy things to 1.8.8.

"Too full" here means that there is not enough space in the inventory to complete the recipe transfer. This happens when there are some items in the crafting table already and I need to move them to the player's inventory to make room for the new recipe. There can be some space in the inventory so it's not 100% "full" but it's still "too full" for what I want to do.

mezz:
Merged here: https://github.com/mezz/JustEnoughItems/commit/9433725dcdd209dc2e399474c14e447db63fcfff

Vexatos:
:+1:

PR #71, opened 2015-12-19
New API interface for custom transfer logic to run when + button is pressed

zerofall:


mezz:
Thank you for the PR, but I would like to do this a little differently. 

`IRecipeCustomTransferHelper` should not extend `IRecipeTransferHelper` since the inventory slots are not relevant. Ideally my implementation of recipe transfer should implement `IRecipeCustomTransferHelper` too, to serve as an example. It also might be called `IRecipeTransferHandler`.

I may need to reorganize `IModRegistry` to keep it from being overrun with these recipe transfer functions.

I also want to expose `IRecipeTransferError` to the API and have `handleTransfer` return that.

mezz:
I'm currently working on some stuff that breaks the API so I should be able to get to this soon, this weekend.

zerofall:
Awesome, thanks. I'll close the PR to let you handle it as you see fit

Issue #105, opened 2016-01-12
API Additions


superckl:
I've been working on an addon for JEI (2.16.2.78) and I've had a few issues with interacting with the mod. I'll list them here. I might add more later as I encounter them. If there is a direct or better way to do any of these, please let me know.
- [ ]  Add a way to retrieve the RecipesGui instance. Currently I'm reflecting `guiEventHandler` from `ProxyCommonClient` (gotten from the `JustEnoughItems` instance) and then reflecting that to get `recipesGui`. I would appreciate if a way was added. (Not necessarily in the API. Mods that are simply adding their recipes will probably not need to access the GUI instance.)
- [ ] Add a way to get `logic` from `RecipesGui`, currently using reflection. (Not necessarily in the API.)
- [ ] Add a way to set the specific page (by index) and category (by instance & index) in `IRecipeGuiLogic`. Currently using (a lot of) reflection.
- [x] The experience gained text in the furnace recipe gui extends past the width specified by the background. This causes the text to sometimes continue past the specified area for the recipe. Not entirely sure what can be done about this, maybe a way to move the text left if needed?

Another more general point - A lot of classes that I could foresee extending have all private fields. It would be useful if the classes that are likely to be extended have protected fields.

superckl:
In response to your comment on #114, I'm going to list out the ways in which I am currently forced to interact with non-API classes.
- Retrieving the RecipesGui (required to open it, no proper abstraction in API) instance requires using `JustEnoughItems`, `ProxyCommonClient`, and `GuiEventHandler`.
- Retrieving RecipeLayouts from JEI properly requires using `Focus`, `Focus.Mode`, and `RecipeGuiLogic`. The needed methods are present in `IRecipeGuiLogic` (which isn't even in the API), but there is no way to get an instance that I know of without referencing `RecipeGuiLogic`.
- I am unable to abstract to `IRecipeLayout` since `IRecipeGuiLogic.getRecipeWidgets` returns a list of `RecipeLayout`.
- Retrieving any specific information from an `IRecipeWrapper` (e.g. `SmeltingRecipe`) requires referencing the wrapper class directly (and reflection in most cases).
- Retrieving `IRecipeTransferHandler` for a container requires referencing `Internal` to get a reference to `RecipeRegistry`. The required method is not present in `IRecipeRegistry`.
- Transferring a recipe should be done with `RecipeTransferUtil`, which is not in the  API.

Hope this helps.

mezz:
[...]

PR #114, opened 2016-01-14
Add features requested in issue #105

superckl:
This commit adds all features request in #105.


mezz:
I appreciate the PR but I'm not going to support using internal methods of the mod, I want to add these features to the API properly.
When I change things in the mod, I need to know what's being used (the API) so I can avoid breaking addons. Stability is extremely important for this project, and it means I will be moving a bit slowly on features requests like this, sorry.

superckl:
Understandable. I'll resort to temporary fixes for issues in my addon until you can support the needed features in such a way.

PR #234, opened 2016-03-31
null pointer exception

ghost:
When the configuration is opened in the main window before the game started, thePlayer is null in Minecraft, and therefore, fmlClientHandler.getClientToServerNetworkManager() is null also.

### DIFF
--- a/src/main/java/mezz/jei/config/SessionData.java
+++ b/src/main/java/mezz/jei/config/SessionData.java
@@ -30,7 +30,7 @@ public class SessionData {
 			FMLClientHandler fmlClientHandler = FMLClientHandler.instance();
 			final NetworkManager networkManager = fmlClientHandler.getClientToServerNetworkManager();
 			if (networkManager == null) { // when opened in main window before a game is started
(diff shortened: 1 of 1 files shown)


mezz:
Thanks. This needs a bit more work, I don't think it's a good idea to return null from here. It would be safer to make that line `worldUid = "none";` rather than returning.

ghost:
Ok, did this, but you'd find a way to handle the "none" value in the methods that use this return value. Null is always easier to check... but the rest is up to you. Regards...

mezz:
Not quite what I meant... I want to set the variable, not return a value there. See how the rest of the method does it. Anyway don't worry about it I can fix it up.

ghost:
this way better? I'd wonder if it is correct to build up a hash code from "none", therefore did I put it all in an extra if tree.

mezz:
There were some other issues that came up besides null here, I fixed them in this commit https://github.com/mezz/JustEnoughItems/commit/ab401cf918412b9693c2e77daf170631bef96e26

PR #258, opened 2016-04-22
Add API jar creation + maven uploading

TehNut:
Creates an extra `.jar` that contains only the API classes and source.

Useful for those who want to add JEI support but do not wish to have an extra mod loaded in the development workspace (or can't for other technical reasons).

http://gfycat.com/UnfortunateFavoriteDungbeetle

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -99,9 +99,17 @@ task sourcesJar(type: Jar) {
 	classifier = 'sources'
 }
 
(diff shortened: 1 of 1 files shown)


mezz:
Is the API properly self-contained now? I remember I had it brokenly requiring JEI internals in the imports before, and I haven't checked it carefully since then.

PR #268, opened 2016-05-08
French translation file for the mod

jorys-paulin:
Hello,
I made the translation file for the french language. So I made the "fr_FR.lang" file located in the same directory as en_US.lang.
I know that many other french players uses the "Canadian french" language file, "fr_CA.lang", but I don't know if I should just duplicate the "fr_FR" one.


mezz:
Thanks! 
I don't think the "fr_CA" file is necessary, even Forestry only has "fr_FR".

PR #346, opened 2016-07-06
Fix for items with both empty and some nbt

P3pp3rF1y:
I noticed that recipes were working for Reliquary's potions that have nbt but no recipe was coming up for empty potion vial that uses the same item, but no nbt.

Looks like the issue is that when it is registering the recipes the stack doesn't have nbt and thus returns null in subtypeinfo and nothing gets added to id. However when user clicks on the item in the JEI list it has empty nbt and thus ":{}" gets added to the id that is searched and it obviously won't find any recipe for that.

Thus the check for no tags that returns null as well. 

This is likely very much a corner case and I fixed that by implementing an nbt interpreter on my side, but may be good for others if they have similar case of an item.

And yeah feel free to change the title to something that makes more sense as it's probably too late for me to come up with a well thought out English sentence.

### DIFF
--- a/src/main/java/mezz/jei/SubtypeRegistry.java
+++ b/src/main/java/mezz/jei/SubtypeRegistry.java
@@ -71,7 +71,7 @@ public class SubtypeRegistry implements ISubtypeRegistry {
 		@Override
 		public String getSubtypeInfo(@Nonnull ItemStack itemStack) {
 			NBTTagCompound nbtTagCompound = itemStack.getTagCompound();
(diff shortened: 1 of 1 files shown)


mezz:
Thanks!
Generally I think empty NBT is bad practice, but of course JEI should handle whatever's thrown at it.

P3pp3rF1y:
Yeah I checked if the empty NBT is on my side, but I am only creating item without any nbt at all so it looks like it's vanilla adding it somewhere along the way

PR #385, opened 2016-08-11
Fix arbitrary item creation and server freeze vulnerability

pau101:
Currently, [PacketRecipeTransfer](https://github.com/mezz/JustEnoughItems/blob/568960886949d3e1ed65bcfd68699f0717a44980/src/main/java/mezz/jei/network/packets/PacketRecipeTransfer.java) sends item stacks from the client to the server to be placed into the crafting grid. This can be exploited with a specially crafted packet to either arbitrarily place items in any container or freeze the server.

This PR resolves this issue by instead sending the slot numbers that contain the item stacks to move for the server to move items from its own slots.

The arbitrary item creation exploit works by sending desired item stacks with a stack size of -1 to trip up the logic into providing item stacks of their max stack size. Items that can't stack will remain a stack size of -1. In addition there is alternate means of producing items that do not use negative stack sizes by transforming actual chests in the player's inventory into chests with a [BlockEntityTag](https://minecraft.gamepedia.com/Player.dat_format#Block_Tags) that specify NBT for items in the chest.

To exploit the server freeze vulnerability the "maxTransfer" argument of PacketRecipeTransfer is set to `true` and the "recipeMap" argument contains strictly negative stack size item stacks. The result is the server becoming stuck in the following loop at [BasicRecipeTransferHelper#removeSetsFromInventory:207-216](https://github.com/mezz/JustEnoughItems/blob/568960886949d3e1ed65bcfd68699f0717a44980/src/main/java/mezz/jei/transfer/BasicRecipeTransferHandler.java#L207-L216)

``` java
while (!requiredCopy.isEmpty() && removeSetsFromInventory(container, requiredCopy, craftingSlots, inventorySlots)) {
  removedSets++;
  Iterator<ItemStack> iterator = requiredCopy.iterator();
  while (iterator.hasNext()) {
    ItemStack stack = iterator.next();
    if (!stack.isStackable() || (stack.stackSize * (removedSets + 1) > stack.getMaxStackSize())) {
      iterator.remove();
    }
  }
}
```

In order to get the client to determine what slot numbers the item stacks that will be transferred belong to I have reworked the [StackHelper](https://github.com/mezz/JustEnoughItems/blob/568960886949d3e1ed65bcfd68699f0717a44980/src/main/java/mezz/jei/util/StackHelper.java) such that determing if a list of item stacks contains any item stack from another list works by seperating the return type from the compare type and parameterizing the assorted "contains" method by return type.

I have created a [concept mod](https://gist.github.com/pau101/ef1e32fe5f8731b57eb007635b61cf0d) which exploits these vulnerabilities. It sends an exploit packet which will fill the currently opened container with the item stack under the cursor when left CTRL is pressed.

### DIFF
--- a/src/main/java/mezz/jei/network/packets/PacketRecipeTransfer.java
+++ b/src/main/java/mezz/jei/network/packets/PacketRecipeTransfer.java
@@ -8,7 +8,6 @@ import java.util.List;
 import java.util.Map;
 
 import net.minecraft.entity.player.EntityPlayer;
-import net.minecraft.item.ItemStack;
 import net.minecraft.network.PacketBuffer;
 
 import mezz.jei.network.IPacketId;
@@ -17,7 +16,7 @@ import mezz.jei.transfer.BasicRecipeTransferHandler;
 
 public class PacketRecipeTransfer extends PacketJEI {
 
-	private Map<Integer, ItemStack> recipeMap;
+	private Map<Integer, Integer> recipeMap;
 	private List<Integer> craftingSlots;
 	private List<Integer> inventorySlots;
 	private boolean maxTransfer;
@@ -26,7 +25,7 @@ public class PacketRecipeTransfer extends PacketJEI {
 
 	}
 
-	public PacketRecipeTransfer(@Nonnull Map<Integer, ItemStack> recipeMap, @Nonnull List<Integer> craftingSlots, @Nonnull List<Integer> inventorySlots, boolean maxTransfer) {
+	public PacketRecipeTransfer(@Nonnull Map<Integer, Integer> recipeMap, @Nonnull List<Integer> craftingSlots, @Nonnull List<Integer> inventorySlots, boolean maxTransfer) {
 		this.recipeMap = recipeMap;
 		this.craftingSlots = craftingSlots;
 		this.inventorySlots = inventorySlots;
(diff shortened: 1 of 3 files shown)


mezz:
Thank you! I will take some time to confirm your findings. 
I hope you enjoyed finding this exploit! I have kept the server-side parts of JEI to a minimum to try to keep the exploitable area very small.

pau101:
I do a lot mod vulnerability discovery and I was just looking through JEI source looking for any potential problems and the client sending item stacks to the server stuck out quite a bit and with a bit of trial and error found the setup for the exploit as in my concept mod.

I don't believe my fix for the issue is the best it could be, but it was the simplest given the existing code. This was my fourth attempt since I would end up redesigning too many things. Ideally the server would do all the logic for transferring the item stacks, however that requires a bit of a reconceptualization of the interface to be present on the server.

mezz:
I need to change the code anyway, the whole thing is quite complicated and I settled on this implementation not because it's particularly good, but because I was able to get it working and iron out duplication and deletion bugs. Any change I make to this requires a ton of testing.

mezz:
Looks good. Thanks again!

mezz:
Also just as a random tip, you can add a "useless" generic type to avoid the casting you were doing:
https://github.com/mezz/JustEnoughItems/commit/8aa5183808b74dfa3f965b93eb27e02d6009bb86#diff-c876a77bd24aa19807415c6847dd19b5L161
yay java

PR #421, opened 2016-09-08
Prevent ForgeCaps from being set with zero size

droidicus:
According to Lex on #minecraftforge, the NBT tag ForgeCaps should never be empty. Forge Universal Buckets were being generated with an empty ForgeCaps tag, and this PR will prevent a ForgeCap tag from being set if it has zero size.

This may be a bandaid for an underlying problem, but it seems to fix the issue I was having with displaying NBT sensitive recipes and using cheated-in items from JEI in these recipes.

### DIFF
--- a/src/main/java/mezz/jei/util/StackHelper.java
+++ b/src/main/java/mezz/jei/util/StackHelper.java
@@ -390,7 +390,7 @@ public class StackHelper implements IStackHelper {
 					nbtTagCompound = new NBTTagCompound();
 				}
 				NBTTagCompound forgeCaps = serializedNbt.getCompoundTag("ForgeCaps");
-				if (0 != forgeCaps.getSize()) { // ForgeCaps should never be empty
+				if (!forgeCaps.hasNoTags()) { // ForgeCaps should never be empty
 					nbtTagCompound.setTag("ForgeCaps", forgeCaps);
 				}
 			}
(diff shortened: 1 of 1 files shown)


src/main/java/mezz/jei/util/StackHelper.java
```diff
@@ -389,7 +389,10 @@ public String getUniqueIdentifierForStack(@Nonnull ItemStack stack, @Nonnull Uid
 				if (nbtTagCompound == null) {
 					nbtTagCompound = new NBTTagCompound();
 				}
-				nbtTagCompound.setTag("ForgeCaps", serializedNbt.getCompoundTag("ForgeCaps"));
+				NBTTagCompound forgeCaps = serializedNbt.getCompoundTag("ForgeCaps");
+				if (0 != forgeCaps.getSize()) { // ForgeCaps should never be empty
```
mezz: Please use `hasNoTags()` instead

mezz:
Thanks! Please fix the line I commented on and I will pull this.

droidicus:
It looks like using hasNoTags() does not work here. The problem is that `NBTTagCompound serializedNbt = stack.serializeNBT()` is returning an empty ForgeCaps tag (among others), so the tag exists, but it is empty (size is zero).

mezz:
Checking `!forgeCaps.hasNoTags()` should be equivalent to `0 != forgeCaps.getSize()`

droidicus:
oh right... '!'... I'll just leave this here...

mezz:
Looks good, thanks!

PR #481, opened 2016-10-01
Update de_DE.lang

ghost:
Corrected and updated german translation to latest version


src/main/resources/assets/jei/lang/de_DE.lang
```diff
@@ -1,64 +1,88 @@
...
+jei.tooltip.transfer=Bewege Gegenstände
+jei.tooltip.recipe.ore.dict=Akzeptiert: %s
+jei.tooltip.item.colors=Farben: %s
+jei.tooltip.shapeless.recipe=Formloses Rezept
+jei.tooltip.cheat.mode=Cheat-Modus
 
 # Error Tooltips
-jei.tooltip.error.recipe.transfer.missing=Fehlende Items
-jei.tooltip.error.recipe.transfer.inventory.full=Inventar ist voll
+jei.tooltip.error.recipe.transfer.missing=Fehlende Gegenstände
```
Vexatos: You shouldn't translate the word "Item"

src/main/resources/assets/jei/lang/de_DE.lang
```diff
@@ -1,64 +1,88 @@
...
 
 # Error Messages
-jei.chat.error.command.too.long=JEI muss auf dem Server installiert sein. um mit diesem Befehl klarzukommen, da er zu lang für Minecraft ist.
+jei.chat.error.command.too.long=JEI muss auf dem Server installiert sein, damit dieser Befehl funktioniert, da er zu lang für Minecraft ist.
 
 # Key Bindings
-key.jei.toggleOverlay=Item-Listen-Overlay (de-)aktivieren
-key.jei.showRecipe=Zeige Item-Rezepte
-key.jei.showUses=Zeige Item-Nutzung
+key.jei.toggleOverlay=Gegenstandsliste de-/aktivieren
```
Vexatos: Same here.

src/main/resources/assets/jei/lang/de_DE.lang
```diff
@@ -1,64 +1,88 @@
...
+key.jei.focusSearch=Suchfeld
+key.jei.showRecipe=Zeige Rezept
+key.jei.showUses=Zeige Nutzung
```
Vexatos: Maybe make this "Verwendung"? Not sure.

src/main/resources/assets/jei/lang/de_DE.lang
```diff
@@ -1,64 +1,88 @@
...
+key.jei.recipeBack=Zeige vorherige Rezeptseite
+key.jei.toggleCheatMode=Gegenstand-Cheating de-/aktivieren
```
Vexatos: Please change this to `(de-)aktivieren`. It's the more proper way to spell it.

Vexatos (changes requested):
Please revert all "Item" -> "Gegenstand" changes, this word shouldn't be translated.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: Why? It's the correct german word for it and its translated this way in vanilla.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: Depends on what exactly is meant ingame. I will have a look...

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: It is inconsistent with mods and feels very clumsy of a word. Vanilla's translation often isn't a good example to base your own work on...

src/main/resources/assets/jei/lang/de_DE.lang
```diff
@@ -21,7 +21,7 @@ jei.tooltip.error.recipe.transfer.no.server=Auf dem Server muss JEI installiert
 jei.chat.error.command.too.long=JEI muss auf dem Server installiert sein, damit dieser Befehl funktioniert, da er zu lang für Minecraft ist.
 
 # Key Bindings
-key.jei.toggleOverlay=Gegenstandsliste de-/aktivieren
+key.jei.toggleOverlay=Gegenstandsliste (de-)aktivieren
```
Vexatos: Changing it only in a single line and not all the others makes little sense.

Vexatos (changes requested):

mezz:
Thanks for the contribution!
I trust Vexatos's judgement here, he has done many mod translations. Let me know when this is resolved and I will pull the changes.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: Then all the other inconsistent mods should be made consistent by translating this word the correct way. If you don't like a german word, it's not a good reason to not translate it, as it's just your opinion.
Vanilla's translation is a good example of a consequent translation, apart from not translating Glowstone, which doesn't make sense, but unlike Redstone and Creeper, which have a "value of brand recognition".
If you feel that German is a "clumsy" language, than simply don't play the game in German.

Vexatos: It is not a clumsy language, it's just this word in particular that makes little sense to be translated in some contexes. In this particular mod it would be fine, but I'd like to keep it consistent with other mods instead.

src/main/resources/assets/jei/lang/de_DE.lang
mezz: This is the hotkey you press when hovering over an item to show its uses in recipes. ("U" by default).

src/main/resources/assets/jei/lang/de_DE.lang
ghost: Ok, can you please tell me the other mods where the context makes the word not fitting?
I know that the german players became familiar with a lot of english words, and in most cases the reason for not translating these words simply was missing space on the screen/gui. But if this is not the case, I see no reason to not translate them, as this degenerates the german language just because of laziness. Best example is germans beginning to conjugate english words with german conjugation rules, e.g. "zähmen" becomes "tamen". I don't want to support this.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: Then "Verwendung" fits better.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: Missing space is one of the most important points, actually, as GUI space is hardcoded in Minecraft. [Here is an example](https://github.com/aidancbrady/Mekanism/blob/7f068231d0089e56e4723fbbf9b026be248b4888/src/main/resources/assets/mekanism/lang/de_DE.lang). I used to have issues running out of GUI space, so keeping things short is quite important sometimes.

Vexatos (approved):

Vexatos:
Now the only thing left is the issue with the word "Item".

src/main/resources/assets/jei/lang/de_DE.lang
ghost: After thinking over it for a while and looking into Mekanism, I partially changed my mind. I still would prefer to use "Gegenstand" within complete sentences where enough space is available, like helping tooltips and so on, but something like a "Itemliste" would be ok. What do you think of this "hybrid mode"?

Vexatos: I'd say keep it consistent. Doing both seems confusing.

ghost: Well... No, I can't go further. What really confuses here is what is resulting from the fact that programmers don't give enough space for other languages. Being constistent in terms of beginning to not translate a word _everywhere_ breaks the logic of a consistent translation completely. Instead, giving enough space would be the best solution in order to make it possible to translate correctly. If the programmer is not willing to do so, than it means he doesn't care about other languages and therefore doesn't want a good localization.
So, please go ahead and make Mekanism and similar mods translation friendly instead of violating a language because of trivialities. It's not impossible.

Vexatos: It is an issue with Minecraft, not any particular mod, but it will never be "fixed" there because it's not an issue that occurs in Vanilla. We just have to live with it.
And I still think that "Item" is a better word and also commonly accepted in German anyway.

src/main/resources/assets/jei/lang/de_DE.lang
```diff
@@ -1,64 +1,88 @@
...
+gui.jei.editMode.description.hide.mod.id=Ausblenden per ModId mit STRG-Shift-Klick.
+gui.jei.editMode.description.show.mod.id=Ausgeblendet per ModId. Einblenden mit STRG-Shift-Klick.
 
 # Recipe Categories
 gui.jei.category.craftingTable=Craftingrezepte
 gui.jei.category.smelting=Schmelzrezepte
-gui.jei.category.smelting.experience=%s XP.
+gui.jei.category.smelting.experience=%s XP
 gui.jei.category.fuel=Kraftstoffrezepte
+gui.jei.category.fuel.smeltCount.single=Schmelzt 1 Gegenstand
```
Vexatos: This has to be "schmilzt".

src/main/resources/assets/jei/lang/de_DE.lang
```diff
@@ -1,64 +1,88 @@
...
-config.jei.title=%MODNAME Konfiguration
+config.jei.title=%MODNAME-Konfiguration
 
 config.jei.mode=Modus
 config.jei.mode.comment=Ändere den Modus, in dem sich JEI befindet
-config.jei.mode.cheatItemsEnabled=Aktiviere Item-Cheating
-config.jei.mode.cheatItemsEnabled.comment=Gibt Items, statt die Rezepte zu zeigen.
-config.jei.mode.editEnabled=Aktiviere Item-Listeneditiermodus
-config.jei.mode.editEnabled.comment=Verstecke und zeige Items durch Klicken in der Item-Liste.
+config.jei.mode.cheatItemsEnabled=Aktiviere Gegenstands-Cheating
```
Vexatos: I do still think that "Item-Cheating" sounds better here.

Vexatos (changes requested):

src/main/resources/assets/jei/lang/de_DE.lang
ghost: The shortage of gui space can be fixed, as the gui is part of the program and programs are programmed by programmers. ;-) The day when the monitors were tiny and the memory could be counted with one hand lies decades behind us. It's just something we have to ask for, or in case of mods, program by ourselves.
There's a lot of what is "commonly accepted" in German nowadays, but only by people who think they can't change anything, as you, or people who don't think at all. But the problem of degenerating language is not unique to German, it's everywhere. I just don't want to support it, as I don't find one word or language "better" than another.
So we're at the beginning: It's not a good reason to not translate it, as it's just your opinion.

Vexatos: Consistency. Consistency is a very good reason to not translate it. Consistency with other mods, some of which have no choice but to not have the word translated due to missing space. There is _nothing_ that can be done about this without major hackery, believe me.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: Something smelts = etwas schmilzt
Somebody smelts (something) = jemand schmelzt (etwas)

Vexatos: http://www.duden.de/rechtschreibung/schmelzen

"schmelzt" is outdated German. It is "schmilzt" in both the transitive and intransitive verb.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: A correct translation would be "Gegenstandsschmummeln" or "Gegenstandsmogeln". This would exactly describe what's happening. But why do even I think that not translating this is better? Because the "Cheat" has been adopted in German for cheating inside a computer game long ago, and it has it's own meaning there which is not congruent with the original meaning in Englisch, where it's not limited to computer games.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: But e.g. Mekanism has user interfaces which are not found in vanilla. How did they made them?

Vexatos: In the end, most mods are based on Minecraft's own GUI and button code as it is so complicated that few modders try to do it themselves.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: The word "Item" has been in various games for years, as well.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: "Schmelzt" is only outdated for the strong form. The examples on the Duden website miss the weak form.
http://www.belleslettres.eu/artikel/schmelzen-erschrecken-hangen.php

src/main/resources/assets/jei/lang/de_DE.lang
ghost: Doesn't sound like "We just have to live with it". Why are people giving up so quick?
Do you have a list of mods where the gui makes problems in German?

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: https://de.wiktionary.org/wiki/schmelzen for another source. [And another](http://de.pons.com/%C3%BCbersetzung?q=schmelzen&l=deen&in=&lf=de). [And another](http://de.langenscheidt.com/deutsch-englisch/schmelzen). There is no difference between the transitive and the intransitive verb unless you try too hard to find one.

The Duden also specifically mentions it is the same for both verbs. 
auswahl_050

Vexatos: The duden also specifically mentions it is the same for both verbs. 
auswahl_050

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: You clearly have never tried writing a GUI in Minecraft. Trust me, it is a lot more ugly than it sounds. Minecraft is a very badly written game.

I obviously do not have a list, but having contributed to over two hundred mod translations, I have been facing space issues dozens of times. One good example is the Thaumcraft Thaumonomicon where I was running out of space on a page multiple times.

Vexatos: I obviously do not have a list, but having contributed to over two hundred mod translations, I have been facing space issues dozens of times.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: Don't you see the difference? The "Item" can be translated without a loss of information, while the "Cheat" can't. There's no german singular form for a single cheat in a computer game. Instead, the german singular forms "Betrug", "Schummelei" etc. all describe a process, which is not "handy", nothing you can touch like a cheat code. That's why the Cheat was adopted. A good example for the other way round is the german word "Angst", which has been adopted to English, but not because they don't like the word "fear" anymore, but to describe a special form of fear they have no own word for.

In a good translation, you always have to look what (meta) information is related.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: Well, then let's have a talk to the Minecraft developers, don't you think?
Quality doesn't come by itself... ;-)

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: This just sounds weird though. It sounds too forced. Like you desperately want to use "Gegenstand" no matter the sound of the word.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: It's not an issue that appears in Vanilla, meaning that it will never be fixed. That's how it works.

Forge, on the other hand, won't fix it because this would require a ground-up rewrite of the font renderer and GUI system.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: None of your mentioned sources conjugate the both forms completely, nor even mention that there should be no difference between the conjugation of both forms.
What the Duden tells is that "schmelzen" is a verb that has a strong and weak form. That is correct. But "schmelzen" is one of few words where the infinitive in both forms is identical. Now, that doesn't mean both forms are conjugated the same way. You just conclude this from what some websites are not mentioning. They are simply incomplete in this case.

Vexatos: auswahl_051

The first being the intransitive meaning, the second being the transitive one.

It is the same.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: In my world, I can ask someone for help in something I think would be useful for me, and there's a possibility that this person will help me even if it would not be useful for him. I call this concept "good will cooperation". But it only works when the one I want the help from is spoken to. It fails 100% when the one is not asked. And what I can tell from experience, often it exposes to be more easy than you thought before, as all thinking before comes from a mind that is alone, while when speaking to others and cooperate, more things become possible.
In short: Don't trust your judgement when you're alone. Talk to each other. ;-)

Vexatos: You have way too much faith in Minecraft. Noone would ever bother to put as much work as this would be into it. Especially not if it only concerns modded Minecraft.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: To value words by their sound is not part of the translation process.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: But "schmelzen" is one of few words where the **infinitive** in both forms is **identical**.
Now, **that doesn't mean both forms are conjugated the same way**.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: It is. A translation that sounds forced or cannot be pronounced without switching the sound of your voice is not as good as one that easily trips off the tongue.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: That's the point: You have no faith.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: No, "schmelzen" is one of those words where the transitive and intransitive versions are identical. That's what all my sources tell me.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: I have already tried too many times to have any. Just deal with how it is and stop thinking it could happen.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: That's right, but that rule doesn't allow to break the translation process by not translating at all. If there's a word you don't like because of how it sounds, but it's the correct translation without alternative words in the same language, than this word is to be used. That's the core of the language concept - it makes the languages differ from each other and is the reason why you are **translating = moving the sound from one language to another**.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: What you say sounds sad. I don't want to be sad.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: They are simply incomplete in this case.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: Translating: Changing text so that it makes sense in another language without changing its meaning too much.

All it has to do is make sense. Noone forces you to translate every single word if the original word makes just as much sense in your target language. Once you realize that, you can focus on making the translation sound nice instead. Translating every word even if it feels forced makes the translation worse than if you keep a few words to make it sound nice.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: I'd rather trust three reputable dictionaries and my own experience than an online magazine.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: Modding Minecraft is all about trying to get around its bad code. That's part of what makes it challenging. Some things, however, you just can not easily get around.

ghost: Where's the challenge in getting easily around? ;-)

Vexatos: > ~~not easily~~ not at all

Just give it up already. This is an issue and it will stay one forever, so just try to keep things short when it matters. And then keep the rest consistent.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: And we are at the beginning again: The german word "Gegenstand" is not worse than the english word "item" - I can't comprehend that it sounds bad or forced or is hard to be read or spoken - it's just your opinion. :-)

Vexatos: Once you get used to translating you start getting a sense for what sounds good and what doesn't. In this case, "Gegenstand" just sounds forced and clumsy compared to the alternative. Noone would mind it staying "Item" as the meaning is exactly equivalent in this case; furthermore, "item" has, as I've said, been commonly accepted in gaming as a word in German, so it also has the additional advantage of being a specifically gaming-related word, which makes it even more fitting.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: I'd rather ask the three reputable dictionaries why their websites don't mention what I learned in school, and if they have changed the rules since I were in school, I would ask them why they have changed the rules without making comprehensible notes on their websites. :-)

src/main/resources/assets/jei/lang/de_DE.lang
ghost: This breaks the simple rule of **first being correct** and **then being consistent**. Giving up to establish reasonable rules means to tolerate errors, and being consistent after tolerating errors means to reproduce them. Being fine with reproducing errors is an attitude that makes things worse instead of better - that's why I'm talking about a "degeneration".

src/main/resources/assets/jei/lang/de_DE.lang
ghost: In case of a translation, the not translated word is only an alternative when there is no translation available. You just make the error of accepting the not translated word as an alternative because of your opinion of how it sounds, while a translation would not result in a loss of information. That's ok for you, but it's nothing you can make a reasonable rule of, as a rule of this sort has to consider all the opinions of the others. You say it's commonly accepted, but it's only accepted by the ones who are fine with making the same error like you. I just don't want to admit that it's ok to "soften" reasonable rules just because a majority of people has given up or is fine with making errors or just doesn't think of it at all. These are just not good reasons to change rules.

ghost: Btw: Why do you think I'm not used to translating? You didn't ask what I'm used to, and just because you have helped translating 200 minecraft mods doesn't mean you were completely correct in all of your decisions.

Vexatos: I did not talk about you specifically at all, I talked about translating in general. Of course my translations aren't always good either; your translation is really good for the most part, some of it is just somewhat... misguided. You underestimate the impact that the sound of a word has on the quality of the translation. "Gegenstands-Cheating" is a prime example of this issue; it forces you to change your tone and voice inside a word, as you have to switch from German to English, two languages with, in this particular case, completely different pronounciations. This makes the whole word hard to pronounce, and reading it out loud forces you to stop in the middle of the word for longer than it should.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: Consistency is much more important than perfect correctness. As an example, I do believe that it is better to have a mod not translated at all if the alternative would be to have it translated only partially (i.e. only a part of the language entries translated). If a translation is not consistent, it becomes confusing because you have two words meaning the same thing in the same mod.

ghost: Two words meaning the same thing is a concept that exists and is used even within a language. Your "false consistency" is just your inflexibility in accepting that. It is you who wants a "perfect correctness", but your correctness is not reasonable.

I suggest the following reasonable rule:
_If there's not enough space in the gui and the programmers are not willing or able to change it, than the word must not be translated. To avoid to be forced to not translating a word everewhere else based on the false concept of error reproducing consistency, the translated and the not translated word are considered as synonyms. Therefore it is ok to have both within the same game/mod._
This would work because the people know the meaning of both words (as you've confirmed). Even you would be able to live with it, just be a little bit flexible instead of forcing that only one word is allowed.

So we're again at the point where we were 4 days ago:
I still would prefer to use "Gegenstand" within complete sentences where enough space is available, like helping tooltips and so on, but something like a "Itemliste" would be ok. What do you think of this "hybrid mode"?

Can you allow synonyms or must it be one word no matter what?

src/main/resources/assets/jei/lang/de_DE.lang
ghost: This is not the result of my translation, it's the result of adopting a word from a foreign language. Accepting to not use german words just because it's hard to pronounce an adopted english word when combined with german words is the beginning of transforming German into English, as this would be a chain reaction.
So if you are ok with adopting words from other languages, you should be ok with how it's pronounced then. If you're not ok with that, than adopting and using it nevertheless and beginning to transform a whole language is what I call misguided. If it doesn't fit, because of how it sounds or whatever, than simply don't adopt it and find a better alternative.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: This has nothing to do with flexibility. It is just intentionally inconsistent. `"Why is it translated in <this> but not <there>?"`
This doesn't have anything to do with synonyms either. Those are an entirely different thing. There is no good synonym for "Gegenstand".

Just use "Item" everywhere. Please.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: I am not transforming a language. There is no word that is not already in use by a big part of native speakers. I have a choice and I choose to use the one that sounds best, that's how you make a good translation.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: But the question has a simple answer:
_It is not translated there because it doesn't fit in the gui and the programmers don't change it. So instead of making it unreadable, it's just not translated._

Why do you have a problem with that? It's reasonable and the people would understand it.

Vexatos: Because in this case you can actually do it even better; "Item" is not just _not translated_, it is also a word that has a specifically gaming-related context in German. It is much more fitting than "Gegenstand" in every sense of the word.

ghost: What you call "intentionally inconsistent" is just your displeasure of changing what you become used to because of giving up. That's a psychological effect that wants to prevent you from admitting that giving up was not the right decision, as this would put into question why you gave up, and that is challenging.

It's all about flexibility here. :-)

Vexatos: This has little to do anymore with the GUI issue. It is very much a more fitting word in every other sense. In German, the word would convey the intended meaning perfectly; "Gegenstand" wouldn't quite as well. Using both words in the same translation isn't flexibility, it's inconsistency.

ghost: >  It is much more fitting than "Gegenstand" in every sense of the word.

No, it is not. It's just your opinion, what you became used to. :-)

ghost: Please explain the "specifically gaming-related context in German" for the word "Item". What is the information the word "Gegenstand" has not? I don't get it.

Vexatos: "Item" is a word that is used in German _only and specifically_ in games. It's nowhere else to be found. This adds additional meaning to the word which makes it fit better into the translation of a game; it makes it consistent across games.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: You're following rules that allow to transform a language just because of opinion. You don't have the choice you want to have. You don't own the language. You should take into account what others think and to what rule softening can lead to. You can do so, but then you should translate for yourself instead of forcing your opinion to others.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: But what's the additional information that would get lost when translating "item" into "Gegenstand"? I have a lot of games where the word was translated, and I never missed any information there. Besides that, I am aware of the fact that I'm playing a game, I don't need special "not-translations" for it to tell me. Words with meta-information like "cheat" are another case, but "item" doesn't belong to these words.

You're going the wrong way, it leads you to transform German into English very slowly because more and more words fall through the grid, and the grid is widened more an more because of "false consistency". The players get used to it and begin to find german words not suitable, they even think the german words are incorrect while they are not. I don't want to say it's an agenda, some think so, I disagree. But I can see the process, I'm observing it for many years now. It's resulting from convenience and laziness, and this simply is no good reason to do what you want to do.

You became used to going the wrong way. Just think of it.

Vexatos: It's not a wrong way. There are words that do not have an exact German equivalent. This one isn't exact either. It is close, but using "Item" just is better in this case because of consistency, it fitting better, and it actually sounding better in some contexes.

ghost: "Gegenstand" is the exact german equivalent of the englisch word "Item" no matter what you think of it. It was the exact equivalent before you where born and will be in the future, as there is no need to change that. Even computer games don't need to change that.

You misuse the concept of consistency to force your not reasonable opinion onto others, what I call "falsce consistency", and I don't want to support it.

Vexatos: Language changes over time, that's how it's always been and that's how it always will be; it's not about a need to change, it just happens. English words appearing in the German language is just a part of that.

wshadow: Well, this sounds like a fun discussion - let me take part in it :)

1) Modern languages are evolving. That's just how languages work. The people that use it, also adjust it to fit the environment they live in. If you don't like this, stick to a dead language.

2) I would also consider "Item" to be a perfectly good choice in this context. There even is a [Wikipedia page](https://de.wikipedia.org/wiki/Item_%28Computerspiel%29) to explain the meaning of the word to those who don't understand it.

3) Having consistent naming is important. I've seen many players get confused by inconsistent naming (i.e. "fluid" vs. "liquid" and "pipe" vs. "tube").

4) Can we just keep "Item"? At least until Mojang fixes the GUI system?

ghost: Well, hello... :-)
Yeah, we can use "Item" in German, I have no "anti-item-ideology" or something like that. But I think it's important to not let german words disappear just to make it consistent at any cost. Synonyms prove that people are capable of using more than one word to describe a thing, so why do you want to erase the german word "Gegenstand" from computer games as if it's something that doesn't belong there?
I suggested the "hybrid mode": e.g. call it "Item-Cheat" and in the tooltip, explain it with "Gibt den Gegenstand, statt das Rezept zu zeigen." In that case, people who don't speak English at all can learn that an "Item" is a "Gegenstand", because they know the latter, and people who already know simply don't bother. I just don't understand why you are bothered by this.
Following your logic you should also erase all the other synonyms, as people could be confused by them too. I don't think that people are that stupid, and if it seems so, e.g. "pipe" vs. "tube", than PLEASE don't let the stupidity win by giving up, but instead teach them so they can learn. Isn't that a better way?

src/main/resources/assets/jei/lang/de_DE.lang
```diff
@@ -1,64 +1,88 @@
...
 
 # Recipe Categories
-gui.jei.category.craftingTable=Craftingrezepte
+gui.jei.category.craftingTable=Crafting-Rezepte
 gui.jei.category.smelting=Schmelzrezepte
-gui.jei.category.smelting.experience=%s XP.
+gui.jei.category.smelting.experience=%s XP
 gui.jei.category.fuel=Kraftstoffrezepte
+gui.jei.category.fuel.smeltCount.single=Schmelzt 1 Item
+gui.jei.category.fuel.smeltCount=Schmelzt %s Items
```
Vexatos: This still needs to be changed.

Vexatos (changes requested):
Apart from the mentioned grammar fix, I am fine with the changes as they are now.

Vexatos:
Using "Gegenstand" in the full sentences, i.e. with proper context, sounds fine enough, I guess.

src/main/resources/assets/jei/lang/de_DE.lang
wshadow: Yes, you should avoid synonyms in this context, but you don't have to erase them all.
Regarding the "pipe"/"tube" example: There may be a reason to use the different word, if "pipes" are not compatible with "tubes". But if they are the same thing, I would keep using the same word everywhere. Because there might be a different mod out there with "tubes" and "pipes" which are not compatible with each other. If I train my users to use the words synonymously, they will be confused if they use that other mod.

Of course most users will very quickly understand how the different mods work and how the items are named. But there will be someone who is going to rage because he just wasted a virtual diamond. I just don't see the advantage of using synonyms here, when you can just as well use the same word every time you talk about the same thing/concept, and avoid confusion.

ghost: Ok, you are absoluty right in that example and I agree with that. But when the context and tooltips tell you that an "Item" and a "Gegenstand" is the same, there can no confusion happen, as it's the opposite case: It tells you what terms can be used as synonyms, as they **already** mean the same.

Example:
config.jei.interface.overlayEnabled=Aktiviere Item-Liste
config.jei.interface.overlayEnabled.comment=Zeige die Liste der Gegenstände neben einem geöffneten Benutzer-Interface.

This tells everybody that a "Item-Liste" is a "Liste der Gegenstände", so a confusion is impossible, and the reader learned this by the way if he did not know before. Your argument that a confusion can happen when one of the terms is used with another meaning doesn't fit, as in the case of "Item" and "Gegenstand", there is equality in the information, they are used for the same thing and will be in the future. I know no game or other context where an "Item" occurs but is not a "Gegenstand", and I can't imagine any context where this could change.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: I asked Mr. Duden himself and will wait for the answer. No need to hurry. ;-)

Vexatos: [Here is an interesting article explaining it](http://www.spiegel.de/kultur/zwiebelfisch/zwiebelfisch-abc-schmelzen-schmilzen-a-317852.html).

> In früheren Zeiten wurde das transitive Verb regelmäßig gebeugt, da hieß es dann zum Beispiel "die Sonne schmelzte das Eis" oder "der Goldschmied hat das Gold geschmelzt". Diese Formen haben sich allerdings nicht durchgesetzt.

I am pretty sure that there is no "Mr. Duden" you could ask that is more reliable than the Duden itself.

ghost: Ok, ich schreib mal auf Deutsch, geht jetzt schneller. Also, wenn da steht
"In früheren Zeiten wurde das **transitive Verb** regelmäßig gebeugt, [...] Diese Formen haben sich allerdings nicht durchgesetzt."
dann bezieht sich das auf das "transitive Verb", wie es da auch steht (ich hab's mal fett gemacht). Die Aussage bezieht sich **nicht** darauf, wie das intransitive Verb gebeugt wird, da das innerhalb der Aussage nicht erwähnt wird. In den Beispielen wird nur die intransitive Beugung im Perfekt aufgezeigt, der Rest fehlt einfach, so wie auch die anderen Seiten diese Information einfach vermissen lassen. Und darum hab ich jetzt einfach mal bei der Duden-Redaktion nachgefragt, denn die können mir bestimmt mehr dazu sagen als irgendwelche unvollständigen Internetseiten.

PS: Nimm das alles nicht zu ernst. Ich diskutiere gerne, denn da kann man viel lernen, wie man korrekte Argumentation und Formulierung aufbaut usw. Mir macht das Spaß. Persönlich ist mir ziemlich egal, ob in Minecraft alles bürokratisch korrekt übersetzt ist oder nicht. ;-)

Vexatos: But... In this context we _are_ talking about the transitive word... [You said so yourself](#discussion_r82498304). This article is perfectly relevant and valid.

ghost: Oh, you're right, I swapped them. The argumentation and examples on the zwiebelfisch website are wrong, then. Seems they unified everything, but I learned: the forms ending on "zt" (2nd and 3rd singular, 2nd plural) are written with an "e" for the transitive form and an "i" for the intransitive form.

Vexatos: Right, the Duden, Langenscheidt, Pons and Spiegel are all wrong. Sure.

Seriously, either you change this now or I will fix it with a later Pull Request, whatever suits you best.

ghost: Here's the list how I learned it:

**Starke Beugung - Präsens Singular**
1. Person: ich schmelze (selbst)
2. Person: du schmilzt (selbst)
3. Person: er/sie/es schmilzt (selbst)

**Starke Beugung - Präsens Plural**
1. Person: wir schmelzen (selbst)
2. Person: ihr schmilzt (selbst)
3. Person: sie schmelzen (selbst)

**Schwache Beugung - Präsens Singular**
1. Person: ich schmelze (das Eis)
2. Person: du schmelzt (das Eis)
3. Person: er/sie/es schmelzt (das Eis)

**Schwache Beugung - Präsens Plural**
1. Person: wir schmelzen (das Eis)
2. Person: ihr schmelzt (das Eis)
3. Person: sie schmelzen (das Eis)

What zwiebelfish says, that there is no "schmilzen", is correct. The rest they say is not how I learned it. But you're right, language is evolving and maybe the people don't want it this way. I will change it as soon as Duden tells me that it's not correct. I couldn't derive it from the mentioned websites - they just claim it without giving logically complete arguments and examples. :-)

Vexatos: But... the Duden does specifically claim it, as explained [here](#discussion_r82498771). This means the entire page applies to both forms of the verb, including the mentioned grammar. This is how you read dictionaries...

Rosethorns:
Why not Objeckt? That's what I came up with, with my Munich boyfriend. It's a shorter than gegenstand so should fit in English UI scale and is an accurate alternative

Edit: anot to an, can't edit git comments from cellphone apparently.

mezz:
Should we escalate this to the [Gesellschaft für deutsche Sprache](http://gfds.de/kontakt/)?

src/main/resources/assets/jei/lang/de_DE.lang
Rosethorns: Wshadow. Just to clarify. Language assimilation is different to language evolution. Do not confuse the two. "this party is hot" > "this party is on fire" > "this party is lit" is language evolution. Language assimilation however is not always as innocent.

Vexatos:
@Rosethorns "Objekt" is something quite different in meaning as it has more than just the wanted connotation.
This issue has already been resolved, I'm only waiting for a grammar fix.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: Yes, they claim it and give examples, but no logically complete arguments why it is different from what I learned. So the situation is as follows: I can simply believe what they say without asking, or I can ask to understand why it is different from what I learned. I choose the latter, and I don't understand why you seem to have a problem with that. Maybe you're one of those who believes everything what is claimed - the school system mainly teaches that, so most people are not <strike>thinking</strike> asking anymore - but I don't want to go in that direction. Just let me ask, it doesn't hurt.

ghost:
@mezz 
The german language has been experimented with a lot in the 20th century, and the experiments still go on. There are powerful "clubs" who connote words with negative meanings in order to make the people think in a special way. I don't want to go into detail here, but the GfdS is one of the instances trying to control language in a political way, meaning they use language against people. In short: They are not the ones who should be asked.

"People who write simple, think simple, and simple thinking people are easy to be controlled."

This "agenda" exists, but from what I know, they have nothing to do with the "anglicization" of the german language, which in my opinion is an "evolutional effect" of global culture transfer and the already mentioned laziness in consistent translating - instead some establish a consistent non-translation for some words, as in the case of "item", which I think is false consistency, but I'm repeating myself.

src/main/resources/assets/jei/lang/de_DE.lang
Vexatos: Currently it's the grammar you learned versus the grammar I learned which three major dictionaries support... Just change it already. Please. As the article stated, "schmelzt" was trying to be introduced but they failed.

mezz:
Don't worry @mankra, I was only joking. My partner has a degree in linguistics, so I have some perspective on this issue. English is extremely lazy about this type of thing so I laugh to myself a little bit, but I also take localization seriously so thank you!

ghost:
@mezz 
I'm not worried, I just spread a little bit knowledge because of too much time. ;-)

ACGaming:
And as we talk, the German language is slowly degrading on Facebook, WhatsApp and numerous other social media platforms.

src/main/resources/assets/jei/lang/de_DE.lang
ghost: I asked, they answered, I changed it. Everything is good. :-)

Vexatos (approved):
Finally!

mezz:
100+ comments and actually came to a conclusion, impressive.

PR #510, opened 2016-10-09
Feature/en au translation

AtomicBlom:
Here's a serious translation. I've split it into two commits so you can see I was being serious this time.


mezz:
Thanks!

PR #526, opened 2016-10-16
minor fix

sokratis12GR:


mezz:
Thank you!

@ELChris414 do you agree with this change?

ttofis:
I tottaly agree, and thank you Sokratis for spotting this mistake that i made! (Σε ευχαριστώ φίλε!)

ttofis:
And mezz just so you understand the fix that he made, and i am really thankful for him doing since i told you, i translated it too quickly, is the gender of the word and also capitalisation

mezz:
Sounds good, thank you both!

PR #527, opened 2016-10-18
Added ArmorPlus in the "Plugin Examples" section

sokratis12GR:
Just wanted to add ArmorPlus' JEI integration in the examples section if it is possible.


mezz:
Thanks but I don't think we need more mod examples now.
My next step is to work on a wiki to show people how to use the API in a more organized way.

Issue #478, opened 2016-09-30
Change ModID to lowercase


mezz:
This will be enforced by FML in Minecraft 1.11

mezz:
moved here https://github.com/mezz/JustEnoughItems/issues/554

PR #528, opened 2016-10-19
Changed the mod ID to lowercase

evieluvsrainbows:
I have changed the mod ID to lowercase since it'll be enforced by Forge in Minecraft 1.11.

### DIFF
--- a/src/main/java/mezz/jei/config/Constants.java
+++ b/src/main/java/mezz/jei/config/Constants.java
@@ -8,7 +8,7 @@ public class Constants {
 	public static final String minecraftModName = "Minecraft";
 
 	// Mod info
-	public static final String MOD_ID = "JEI";
+	public static final String MOD_ID = "jei";
 	public static final String NAME = "Just Enough Items";
 	public static final String VERSION = "@VERSION@";
 
--- a/src/main/resources/mcmod.info
+++ b/src/main/resources/mcmod.info
@@ -1,6 +1,6 @@
 [
 {
-  "modid": "JEI",
+  "modid": "jei",
   "name": "Just Enough Items",
   "description": "Simple recipe and item helper.",
   "version": "${version}",


mezz:
Thanks, but I can't change it now or else lots of mod compatibility will break.
See https://github.com/mezz/JustEnoughItems/issues/478

evieluvsrainbows:
@mezz can't a 1.11 branch be created that'll keep 1.11 changes separate from the 1.10 branch, while retaining full compatibility with mods?

evieluvsrainbows:
Wait...I'm going to test this change with my mods (almost all of them take advantage of JEI), and see if any mods break.

**Edit**: I have to setup the environment first, haha.

mezz:
Sure, a 1.11 branch could work, but there's not much point to do it now when Forge is not done.

The issue is with mods calling stuff like `@Optional`, or checking `Loader.isModLoaded`. The change in modId will break them. Most mods that just use the JEI API to add their recipes and such will not break.

evieluvsrainbows:
Ah, so `Loader.isModLoaded` and `@Optional` are case-sensitive then?

mezz:
Yes, definitely.

evieluvsrainbows:
@mezz Damn, that's kinda sucky.

However, small thought on my mind:

I don't think it would require that much effort from the devs to change their code that checks for JEI. If they change the mod ID they search for, compatibility wouldn't be an issue. It'll be a breaking change, but it might be worth it for the future-proofing before 1.11 actually comes out. (won't happen for another little while I don't think, but then ofc we have to wait for Forge to update.)

mezz:
It's best to make breaking changes at the same time as 1.11, modders will have other breaking changes to deal with at the same time and will expect it. Also, Forge 1.11 will probably change the way `@Optional` and `Loader.isModLoaded` work so that they only check lowercase, since all modIds will be enforced as lowercase then.

evieluvsrainbows:
Ah, yeah. That's fair enough. :)

P.S. I'm still going to check my modded instance with the lowercased mod ID change and see if any mods break (if any).

mezz:
Check mods that have deeper integration, like EnderIO which has a recipe button on its machine guis.

evieluvsrainbows:
@mezz You were right...A mod did break (Chisels & Bits). But apparently that's because it was checking explicitly for version 11.15.0.1697 or above. Because my version is just 3.12.8, that probably broke the version checking.

mezz:
11.15.0.1697 is a Forge version, I think you're seeing something else.

evieluvsrainbows:
image

Nope, it says "jei 11.15.0.1697"

mezz:
Looks like Chisels and Bits has a broken JEI dependency info, they use the lowercase ID and have an invalid version.
https://github.com/AlgorithmX2/Chisels-and-Bits/blob/1.10.2/src/main/java/mod/chiselsandbits/core/ChiselsAndBits.java#L60

mezz:
reported here: https://github.com/AlgorithmX2/Chisels-and-Bits/issues/155

evieluvsrainbows:
@mezz Alright. By the way, the way Chisels and Bit's modID checking works is that it doesn't check explicitly for modIDs, it just searches all modIDs so that way mods that change their modIDs don't break.

Source: https://github.com/AlgorithmX2/Chisels-and-Bits/blob/697d01e03068333f14bb324a9cd3c0948ef96eb8/src/main/java/mod/chiselsandbits/integration/Integration.java#L42-L50

mezz:
But basically you never know what will break. 
It's extremely important that JEI never makes breaking changes except on new Minecraft versions, because so many mods interact with it. If I make a change that requires just a minor fix, it will still take hundreds of devs a long time to get that information and make the fix. In the meantime, mods that have updated can't work in the same pack with mods that have not updated, so the entire modded ecosystem is divided until everyone can update.

evieluvsrainbows:
@mezz Ah, okay. Fair enough. And I think that was the wrong file, but that was the only file in the Chisels and Bits repo that uses that isModLoaded check.

AEnterprise:
maybe a solution would be to temp change those checks in forge to also check for the lowercase version to make the transition easier?

evieluvsrainbows:
@AEnterprise Possibly.

mezz:
I think they plan to do that when 1.11 comes.

PR #561, opened 2016-11-02
Small refactoring of the bar calculations

RainWarrior:


mezz:
Tested and looks great, thanks fry!

PR #565, opened 2016-11-03
Fix texture bug in new tabs

### LINKED ISSUE #564: [3.13.1.343] Missing Textures
I updated to this version today and all JEI textures are missing. I looked in the code and in the textures folder and found out that the 2 textures are called "recipeBackground2.png" and "recipeBackgroundTall2.png" instead of "recipeBackground.png" and "recipeBackgroundTall.png" if I would rename these textures in the jar file (remove the "2" from the end of their name) they work just fine.

JEI: 3.13.1.343
NEI: 2.1.2.162
Forge: 12.18.2.2118

Screenshot:

miuirussia:
Fixes #564


mezz:
Thanks, I already had this working locally so I just commited it before I saw this PR.

PR #596, opened 2016-11-09
Allow configuration of mod name tooltip

### LINKED ISSUE #590: Mod Name Config
- It would be nice to see a toggle for the mod name for users who don't want it shown.
- And maybe a config for specifying the colour/style formatting of the modname too.

###### I didn't add [Suggestion] to the title because I assume you'd rather use a label.

TehNut:
Would close #590. If `%s` is not found in the string, it will fallback to the default formatting (Blue with italics)

https://streamable.com/t50a


ChloeDawn:
Technically in order to resolve https://github.com/mezz/JustEnoughItems/issues/590 you'd also need to add a toggle to disable the mod name tooltip entirely. Either that or allow the string to be empty like you have in HWYLA.

mezz:
Thanks, I'll modify this a bit to allow disabling the tooltip.

PR #663, opened 2016-12-13
Fix edge case of fluid containers with subtypes being mashed together

HenryLoenwind:
* re https://github.com/SleepyTrousers/EnderIO/issues/3906


mezz:
Thanks!

PR #683, opened 2017-01-01
Replace ItemBlacklist with IngredientBlacklist

WayOfTheQway:
- Added IngredientBlacklist api and implementation
- Deprecated ItemBlacklist api
- Replaced ItemBlacklist implementation with wrapper around
IngredientBlacklist
- Replaced usage of ItemBlacklist internally with IngredientBlacklist

I need a way in my own mod to hide anything that can be shown in the JEI window, and I think it just makes sense in the api for it to be possible. It's a relatively small change, and it (should hopefully) be backward compatible, since the old ItemBlacklist still exists.

Pretty please, with a cherry on top? :D

P.S. You might wanna run your formatter or whatever your IDE calls it to replace spaces with tabs, move curly brackets, etc. Sorry.


WayOfTheQway:
Alright I fixed the two actual errors.
The unused import it shows now is referenced in a comment.
The NullPointerException is just code straight from the old implementation, so I don't know if that should be changed or not.
Also I'm realizing that this api class probably belongs in the ingredients package. Whoops. I guess I'll let you decide if you'd like to move it there.

mezz:
Thanks this makes sense.

>P.S. You might wanna run your formatter or whatever your IDE calls it to replace spaces with tabs, move curly brackets, etc. Sorry.

Please do that yourself... your style looks fine except for where you use spaces instead of tabs.

>The NullPointerException

I've been switching to using `Preconditions.checkNotNull(parameter, "error message");` for parameter checking, and it passes the style check.

>this api class probably belongs in the ingredients package

That's a good idea, please move it there.

mezz:
Merged here https://github.com/mezz/JustEnoughItems/commit/c9e5444239b2df54346513a6d18f1c1d70731bd6

PR #684, opened 2017-01-01
Replace ItemBlacklist with IngredientBlacklist

WayOfTheQway:
- Added IngredientBlacklist api and implementation
- Deprecated ItemBlacklist api
- Replaced ItemBlacklist implementation with wrapper around
IngredientBlacklist
- Replaced usage of ItemBlacklist internally with IngredientBlacklist

I did it in 1.10 first, since that's the version I primarily use, but I copied that over to 1.11 seeing as that's the newest version. It should work, I looked over everything and changed the error checking stuff, etc.

P.S. You might wanna run your formatter or whatever your IDE calls it to replace spaces with tabs, move curly brackets, etc. Sorry.


WayOfTheQway:
If there's some way to PR two branches at once you should totally tell me, because I have no idea about it.

WayOfTheQway:
The unused import codacy shows is referenced in a comment. :|
Also I'm realizing that this api class probably belongs in the ingredients package. Whoops. I guess I'll let you decide if you'd like to move it there.

mezz:
Merged here https://github.com/mezz/JustEnoughItems/commit/9f3dd11ec17055a3f18887ccd7a75f690df0152c

Issue #689, opened 2017-01-04
if you click on the recipies Button with items in hand they dissappear from inventory and are thrown when you close your inventory


bobis32:
Thank you for reporting!

For suggestions:
 * Please do not suggest features from NEI. [See this spreadsheet](https://goo.gl/MQ0f3R) before requesting.

For issues:
 * Try the latest version, it may have fixed the issue already.
 * If you have NEI or FEI installed, please try without it before reporting.
 * If you crashed, please paste the crash log to [gist](https://gist.github.com/) and link it here.

pau101:
Woah... Just noticed I submitted my PR #690 to fix this after discovering this bug myself, 16 minutes after you submitted this without knowing you had. I imagine this will be resolved soon.

PR #690, opened 2017-01-04
Fix cursor held item and player inventory craft items desyncing when opening RecipesGui

pau101:
This PR changes RecipesGui to not call GuiScreen#onGuiClosed() on the current screen if it is a GuiContainer when itself is opened.

I have analyzed a sample of mod GuiContainers and I am confident that there will not be unintended consequences of this change within conventional mods so long as it is guaranteed RecipesGui will return to the parentScreen eventually.

### DIFF
--- a/src/main/java/mezz/jei/gui/ItemListOverlayInternal.java
+++ b/src/main/java/mezz/jei/gui/ItemListOverlayInternal.java
@@ -11,6 +11,7 @@ import java.util.Set;
 
 import com.google.common.collect.ImmutableList;
 import mezz.jei.Internal;
+import mezz.jei.JeiRuntime;
 import mezz.jei.JustEnoughItems;
 import mezz.jei.api.gui.IAdvancedGuiHandler;
 import mezz.jei.api.gui.IDrawable;
@@ -22,6 +23,7 @@ import mezz.jei.gui.ingredients.GuiIngredientFast;
 import mezz.jei.gui.ingredients.GuiIngredientFastList;
 import mezz.jei.gui.ingredients.GuiItemStackGroup;
 import mezz.jei.gui.ingredients.IIngredientListElement;
+import mezz.jei.gui.recipes.RecipesGui;
 import mezz.jei.input.ClickedIngredient;
 import mezz.jei.input.GuiTextFieldFilter;
 import mezz.jei.input.IClickedIngredient;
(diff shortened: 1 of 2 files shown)


mezz:
Thanks for the contribution!
It's not perfectly clear to me how this fixes the problem, or where all this logic came from.
Can you explain a bit?

pau101:
When GuiContainer#onGuiClosed() is called it will call Container#onContainerClosed(EntityPlayer) which handles dropping the item held by the cursor. When the player inventory is open the ContainerPlayer#onContainerClosed() handles dropping the items the player has in the 2x2 crafting grid.

The logic that has been put in place is a recreation of Minecraft#displayGuiScreen(GuiScreen) except without calling #onGuiClosed() _(except for when the GuiOpenEvent would change what GUI should open)_.

mezz:
Ok that makes sense.
I think the cleanest way to write this is to make it a helper function with a javadoc comment at the top that links to `Minecraft#displayGuiScreen(GuiScreen)` and explains what it is doing. That way I will be able to properly maintain and debug it if Minecraft's method changes.

mezz:
I will go ahead the pull this now and tweak it myself, since it will eliminate a common frustration people run into.

mezz:
Unfortunately I ran into a couple edge cases that make me wary of this kind of change.

- While holding an item in cheat mode and then looking at recipes, clicking an item in the item list will delete the item held.
- While crafting something in the player's gui and then opening and closing the JEI in-game config, the items still disappear and are thrown then the gui closes.

I am not completely sure that modded guis will respond well to this either.

pau101:
The first issue can be mitigated by having RecipesGui render the player's held item and the second is expected because currently displaying the JEIModConfigGui is done regularly in ItemListOverlayInternal#handleMouseClickedButtons.

Until a mod which doesn't behave as intended with this change is found I think it is a worth this improvement.

mezz:
I am worried that a bad mod interaction would enable players to duplicate items.

pau101:
If opening a new Gui while not closing a theoretical mod Gui would enable item duplication that would be vulnerability in that mod and not created by JEI. Same could be said for any mod's _client side_ logic happening to manifest another mod's item duplication vulnerability.

pau101:
I figured not allowing cheat mode item deletion if the RecipesGui is open is the best solution for accidental deletion as opposed to my initial idea, rendering the held item stack.

There is one minor annoyance where if there is an item held by the cursor it could be unintentionally dropped when the JEI config is opened and then closed because mousedown activates the "Done" button and mouseup triggers the item drop. Working around that would be a bit messy and I don't believe it warrants it.

mezz:
Looks better, and works in most of the cases I could think up.
Dropping an item when opening the config is not so bad, I think it will very very rarely come up.

One small issue I found: the item deletion is disabled but it still shows the tooltip.
Can you fix it so that the "Delete" tooltip doesn't show up when you hover over the ingredient list?

mezz:
Thanks, looks good!

pau101:
#689 can also be closed from this.

PR #712, opened 2017-01-28
[1.10] Add a way to add ItemStacks to JEI during runtime (cherry-pick)

Vexatos:
This is a partial cherry-pick from commit https://github.com/mezz/JustEnoughItems/commit/d18a76f3683971b0fea06dea1fc91d56a96c0bd2 for the 1.11 branch It appears to work just fine.

Weird that you have method arguments marked `@Nullable` and then immediately check for not-null...

The deprecated `ItemRegistry` still uses immutable collections, for compatibility.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,4 +5,4 @@ curse_project_id=238222
 
 version_major=3
 version_minor=14
-version_patch=3
+version_patch=4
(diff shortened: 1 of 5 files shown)


mezz:
The API interface is `@Nonnull` and the implementation is overriding it to `@Nullable` so that the IDE does not complain when I null-check it.  
In 1.11 I worked around that more cleanly by using `Preconditions.checkNotNull` instead.

mezz:
You changed too much, can you try to make this as small a change as possible? 
I'm not really comfortable with the breaking API changes in particular, but the other code changes make it more difficult to verify that this will work as intended on 1.10.

PR #732, opened 2017-02-12
Add support for anvil recipes

gigaherz:
> [11:52] (gigaherz): morning ppl
> [11:52] (gigaherz): does JEI come with a template for anvil recipes?
> [11:53] (@mezz): nope, it should though. you can PR if you are feeling motivated

So here it is.

I decided not to attempt generating recipe wrapper instances for the vanilla book enchanting and repair, due to how those are hardcoded:
- For books, I'd need to test each possible item against each possible enchantment, and see if they are allowed. This is somewhat doable, but it seems ugly.
- For repair, Minecraft calls `Item#getIsRepairable(ItemStack,ItemStack)`, with the contents of the anvil slots, and returns true if the material matches the tool. There is no way to figure out the right combinations besides a a nested loop and matching all items against all items.

I'm open to suggestions.

### DIFF
--- a/src/main/java/mezz/jei/api/IModRegistry.java
+++ b/src/main/java/mezz/jei/api/IModRegistry.java
@@ -96,6 +96,7 @@ public interface IModRegistry {
 	 * @param leftInput The itemStack placed on the left slot.
 	 * @param rightInputs The itemStack(s) placed on the right slot.
 	 * @param outputs The resulting itemStack(s).
+	 * @since JEI 4.2.6
 	 */
 	void addAnvilRecipe(ItemStack leftInput, List<ItemStack> rightInputs, List<ItemStack> outputs);
 
--- a/src/main/java/mezz/jei/api/IModRegistry.java
+++ b/src/main/java/mezz/jei/api/IModRegistry.java
@@ -94,15 +94,12 @@ public interface IModRegistry {
 	/**
 	 * Adds an anvil recipe for the given inputs and output.
 	 * @param leftInput The itemStack placed on the left slot.
-	 * @param rightInput The itemStack placed on the right slot.
-	 * @param output The resulting itemStack.
+	 * @param rightInputs The itemStack(s) placed on the right slot.
+	 * @param outputs The resulting itemStack(s).
 	 * @param levelsCost The cost, in experience levels, that will be required in order to pick up the item.
 	 *                   Optional: If not specified, the cost text will not display on the recipe.
 	 */
-	void addAnvilRecipe(ItemStack leftInput, ItemStack rightInput, ItemStack output, int levelsCost);
-	void addAnvilRecipe(ItemStack leftInput, ItemStack rightInput, ItemStack output);
-	void addAnvilRecipe(ItemStack leftInput, List<ItemStack> rightInput, List<ItemStack> output, int levelsCost);
-	void addAnvilRecipe(ItemStack leftInput, List<ItemStack> rightInput, List<ItemStack> output);
+	void addAnvilRecipe(ItemStack leftInput, List<ItemStack> rightInputs, List<ItemStack> outputs, int levelsCost);
 
 	/**
 	 * Get the registry for setting up recipe transfer.
--- a/src/main/java/mezz/jei/api/IModRegistry.java
+++ b/src/main/java/mezz/jei/api/IModRegistry.java
@@ -96,9 +96,8 @@ public interface IModRegistry {
 	 * @param leftInput The itemStack placed on the left slot.
 	 * @param rightInput The itemStack placed on the right slot.
 	 * @param output The resulting itemStack.
-	 * @param levelsCost Optional.
-	 *                   The cost, in experience levels, that will be required in order to pick up the item.
-	 *                   If not used
+	 * @param levelsCost The cost, in experience levels, that will be required in order to pick up the item.
+	 *                   Optional: If not specified, the cost text will not display on the recipe.
 	 */
 	void addAnvilRecipe(ItemStack leftInput, ItemStack rightInput, ItemStack output, int levelsCost);
 	void addAnvilRecipe(ItemStack leftInput, ItemStack rightInput, ItemStack output);
--- a/src/main/java/mezz/jei/api/IModRegistry.java
+++ b/src/main/java/mezz/jei/api/IModRegistry.java
@@ -101,6 +101,8 @@ public interface IModRegistry {
 	 */
 	void addAnvilRecipe(ItemStack leftInput, ItemStack rightInput, ItemStack output, int levelsCost);
 	void addAnvilRecipe(ItemStack leftInput, ItemStack rightInput, ItemStack output);
+	void addAnvilRecipe(ItemStack leftInput, List<ItemStack> rightInput, List<ItemStack> output, int levelsCost);
+	void addAnvilRecipe(ItemStack leftInput, List<ItemStack> rightInput, List<ItemStack> output);
 
 	/**
 	 * Get the registry for setting up recipe transfer.
--- a/src/main/java/mezz/jei/api/IModRegistry.java
+++ b/src/main/java/mezz/jei/api/IModRegistry.java
@@ -92,6 +92,18 @@ public interface IModRegistry {
 	void addDescription(List<ItemStack> itemStacks, String... descriptionKeys);
 
 	/**
+	 * Adds an anvil recipe for the given inputs and output.
+	 * @param leftInput The itemStack placed on the left slot.
+	 * @param rightInput The itemStack placed on the right slot.
+	 * @param output The resulting itemStack.
+	 * @param levelsCost Optional.
+	 *                   The cost, in experience levels, that will be required in order to pick up the item.
+	 *                   If not used
+	 */
+	void addAnvilRecipe(ItemStack leftInput, ItemStack rightInput, ItemStack output, int levelsCost);
+	void addAnvilRecipe(ItemStack leftInput, ItemStack rightInput, ItemStack output);
+
+	/**
 	 * Get the registry for setting up recipe transfer.
 	 */
 	IRecipeTransferRegistry getRecipeTransferRegistry();
--- a/src/main/java/mezz/jei/api/IModRegistry.java
+++ b/src/main/java/mezz/jei/api/IModRegistry.java
@@ -96,10 +96,8 @@ public interface IModRegistry {
 	 * @param leftInput The itemStack placed on the left slot.
 	 * @param rightInputs The itemStack(s) placed on the right slot.
 	 * @param outputs The resulting itemStack(s).
-	 * @param levelsCost The cost, in experience levels, that will be required in order to pick up the item.
-	 *                   Optional: If not specified, the cost text will not display on the recipe.
 	 */
-	void addAnvilRecipe(ItemStack leftInput, List<ItemStack> rightInputs, List<ItemStack> outputs, int levelsCost);
+	void addAnvilRecipe(ItemStack leftInput, List<ItemStack> rightInputs, List<ItemStack> outputs);
 
 	/**
 	 * Get the registry for setting up recipe transfer.
--- a/src/main/java/mezz/jei/plugins/vanilla/VanillaPlugin.java
+++ b/src/main/java/mezz/jei/plugins/vanilla/VanillaPlugin.java
@@ -141,7 +141,7 @@ public class VanillaPlugin extends BlankModPlugin {
 		recipeTransferRegistry.addRecipeTransferHandler(ContainerFurnace.class, VanillaRecipeCategoryUid.SMELTING, 0, 1, 3, 36);
 		recipeTransferRegistry.addRecipeTransferHandler(ContainerFurnace.class, VanillaRecipeCategoryUid.FUEL, 1, 1, 3, 36);
 		recipeTransferRegistry.addRecipeTransferHandler(ContainerBrewingStand.class, VanillaRecipeCategoryUid.BREWING, 0, 4, 5, 36);
-		recipeTransferRegistry.addRecipeTransferHandler(ContainerRepair.class, VanillaRecipeCategoryUid.ANVIL, 0, 2, 3, 4 * 9);
+		recipeTransferRegistry.addRecipeTransferHandler(ContainerRepair.class, VanillaRecipeCategoryUid.ANVIL, 0, 2, 3, 36);
 
 		registry.addRecipeCategoryCraftingItem(new ItemStack(Blocks.CRAFTING_TABLE), VanillaRecipeCategoryUid.CRAFTING);
 		registry.addRecipeCategoryCraftingItem(new ItemStack(Blocks.FURNACE), VanillaRecipeCategoryUid.SMELTING, VanillaRecipeCategoryUid.FUEL);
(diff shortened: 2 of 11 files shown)


src/main/java/mezz/jei/plugins/vanilla/anvil/AnvilRecipeCategory.java
```diff
@@ -0,0 +1,50 @@
...
+import mezz.jei.api.recipe.BlankRecipeCategory;
+import mezz.jei.api.recipe.VanillaRecipeCategoryUid;
+import net.minecraft.init.Blocks;
+import net.minecraft.util.ResourceLocation;
+
+import javax.annotation.Nonnull;
+
+public class AnvilRecipeCategory extends BlankRecipeCategory<AnvilRecipeWrapper>
+{
+    @Nonnull
```
liach: Just use `@MethodsNonnullByDefault` and `@ParametersAreNonnullByDefault` in package info file.

src/main/java/mezz/jei/plugins/vanilla/anvil/AnvilRecipeCategory.java
```diff
@@ -0,0 +1,50 @@
...
+import mezz.jei.api.ingredients.IIngredients;
+import mezz.jei.api.recipe.BlankRecipeCategory;
+import mezz.jei.api.recipe.VanillaRecipeCategoryUid;
+import net.minecraft.init.Blocks;
+import net.minecraft.util.ResourceLocation;
+
+import javax.annotation.Nonnull;
+
+public class AnvilRecipeCategory extends BlankRecipeCategory<AnvilRecipeWrapper>
+{
```
liach: The code style...

gigaherz: Yep, I noticed that JEI uses the wrong brace location after I clicked to create the PR.

src/main/java/mezz/jei/plugins/vanilla/anvil/AnvilRecipeCategory.java
gigaherz: Right, I copied from the wrong example. ;P

mezz:
Looks good!
This will need a few things to move on to the next step:
 * We need a simple API to add recipes, similar to `mezz.jei.api.IModRegistry#addDescription()`
 * Manually add vanilla repair recipes using the API, an itemstack and its repair materials manually. Dynamic generation is too slow here unfortunately. Vanilla only has like 5 item classes that override `getIsRepairable` so it shouldn't be too bad.
 * See how JER adds enchantment recipes, maybe they found a solution? If it's very inefficient, we'll have to think more about it or just leave enchantments out for now.
 * Review from @HenryLoenwind since EnderIO has anvil recipes implemented already.

HenryLoenwind:
Looks basically the same as our anvil recipes. I think I could adapt them to use this instead in about 5 minutes. :+1:

way2muchnoise:
JER just tests if an enchantment can be applied to an ItemStack ([see entry point](https://github.com/way2muchnoise/JustEnoughResources/blob/master/src/main/java/jeresources/jei/enchantment/EnchantmentMaker.java)), I don't know if it is the most efficient way tho. ([actual checking](https://github.com/way2muchnoise/JustEnoughResources/blob/master/src/main/java/jeresources/registry/EnchantmentRegistry.java#L28)). I guess I could look into it again and see if I can find some thing.

HenryLoenwind:
"codacy"? Ok, I'll remember that as "junk to avoid"...

gigaherz:
I'm thinking of adding a 5th parameter to the anvil recipes, so that the recipe can show like, "required: 2+", for cases where the levels cost is variable. Opinions?

liach:
@gigaherz I think that is not defined in a recipe; it is determined by inputs and are calculatable.

HenryLoenwind:
Yes, I think subclassing is the better way here. Maybe split drawInfo(), e.g. in line 40, to make subclassing easier.

gigaherz:
Does this look right for the repair recipes?

Example Recipes

The damage numbers are semi-hardcoded (25%, 50%, 75% of max) right now. I don't know if there's a way to show different damage levels and their respective output dynamically.

gigaherz:
The repair recipes are now implemented. So far as I can tell, I added all the repairable items and their materials, but if there's something I missed, I'd like to know. ;P

As for enchanting, iterating through *all items* and testing each one against *all enchants* seems a bit overkill. It's not hard to implement, but I imagine it would take a rather long time on a big modpack? Are there any numbers from JER regarding that, @way2muchnoise?

gigaherz:
I did a quick test, for vanilla recipes (786 ingredients, 30 enchantments with varying number of levels):
```
[21:08:42] [Client thread/INFO]: [STDOUT]: Registering repair recipes took 6581 microseconds
[21:08:42] [Client thread/INFO]: [STDOUT]: Registering repair recipes took 24656 microseconds
```
So it's not too bad. Event with 10x the items, and 2x the enchantments, it would still be less than half a second.

gigaherz:
Right now, the enchantment recipes show one by one, including separate entries for each book level. Is there some way to show all book levels as different (rotating) alternatives of the same recipe?

mezz:
@HenryLoenwind I tested out codacy to see if it would be usable for style-checking Forge, but I haven't had much time to configure it properly so it's really strict. It may still be useful.

@gigaherz looks good, that doesn't take as long as I thought.
For rotating recipes, just make an input list of (level 1, 2, 3) and output of applied (level 1, 2, 3) and they will rotate together. It's good enough for this purpose.
For the repair recipes, I think you should only show the recipe for (max damaged item -> somewhat repaired item). That way it's easy to see how much is repaired without having recipe spam.

HenryLoenwind:
Be careful with the enchantments, Ender IO has one that can be applied to any item (soulbound).

(semi OT) @mezz, I wasn't commenting on it's strictness, but on its dumbness. Rules that check the signature of a `@Override`ing method do not make any sense in the first place. I take having such rules as a sign for "we don't care but want the biggest number of rules and findings to impress those clueless company buyers and execs so they buy our software which is not even half as good as the free/open source alternatives". I get to work with that kind of !%&$ software at work much too often.

gigaherz:
@HenryLoenwind This has changed in 1.11, it seems. ALL means "all of the above" now, instead of literally "all the items". <del>There doesn't seem to be any actual way to have enchantments that can be applied to all items.</del>
EDIT: I was wrong, there's still `EnumHelper.addEnchantmentType` which can help. XD

HenryLoenwind:
~~So canApplyAtEnchantingTable() was moved from the enchantment to the item so we cannot override it anymore? What's the point in that other than annoying modders?~~

Edit: I looked at the call stack, it is still routed through the enchantment. So still full control for the enchantment by returning true from canApply().

Thx, for the heads-up. ~~Expect people to ASM their hooks in...~~

gigaherz:
I added a custom enchantment, set to apply to *all* items.
```
[20:54:34] [Client thread/INFO] [jei]: Registered vanilla repair recipes in 1 ms
[20:54:34] [Client thread/INFO] [jei]: Registered enchantment recipes in 33 ms
```
in contrast, without that one enchant, I get
```
[20:57:16] [Client thread/INFO] [jei]: Registered vanilla repair recipes in 1 ms
[20:57:16] [Client thread/INFO] [jei]: Registered enchantment recipes in 17 ms
```
So it it takes about as long to register one "all items" enchantment, as it takes to register all the other ones. On the upside, it's in the milliseconds, so even with 10x the items, and 2x the enchants, that's still within a second. :)

HenryLoenwind:
Try one that returns true from canApply() instead.

gigaherz:
@HenryLoenwind Oh I was talking about EnumEnchantmentType.ALL and how it doesn't mean "ALL" anymore.
I mean, yes, if you override canApply, it works. ;P

HenryLoenwind:
Yeah, I know. I skipped a step and looked up what to do to make an enchantment that is applicable to all items again. There are 2 ways: (a) ```canApply(ItemStack stack) { return true; }``` and (b) 
```
EnumHelper.addEnchantmentType("REALLY_ALL", new Predicate<Item>() {
    @Override
    public boolean apply(@Nullable Item input) {
      return true;
    }```

File: src/main/java/mezz/jei/api/IModRegistry.java [hunk 3]
```diff
@@ -92,6 +92,19 @@
...
+	 * @param leftInput The itemStack placed on the left slot.
+	 * @param rightInput The itemStack placed on the right slot.
+	 * @param output The resulting itemStack.
+	 * @param levelsCost The cost, in experience levels, that will be required in order to pick up the item.
+	 *                   Optional: If not specified, the cost text will not display on the recipe.
+	 */
+	void addAnvilRecipe(ItemStack leftInput, ItemStack rightInput, ItemStack output, int levelsCost);
+	void addAnvilRecipe(ItemStack leftInput, ItemStack rightInput, ItemStack output);
+	void addAnvilRecipe(ItemStack leftInput, List<ItemStack> rightInput, List<ItemStack> output, int levelsCost);
+	void addAnvilRecipe(ItemStack leftInput, List<ItemStack> rightInput, List<ItemStack> output);
```
mezz: This should be one method only, not four. Just keep
```java
void addAnvilRecipe(ItemStack leftInput, List<ItemStack> rightInput, List<ItemStack> output, int levelsCost);
```
I don't think hiding the level cost is valid, if there is no cost they can put `0`.

src/main/java/mezz/jei/plugins/vanilla/anvil/AnvilRecipeMaker.java
```diff
@@ -0,0 +1,127 @@
...
+		ItemStack repairStone = new ItemStack(Blocks.COBBLESTONE);
+		ItemStack repairIron = new ItemStack(Items.IRON_INGOT);
+		ItemStack repairGold = new ItemStack(Items.GOLD_INGOT);
+		ItemStack repairDiamond = new ItemStack(Items.DIAMOND);
+		ItemStack repairLeather = new ItemStack(Items.LEATHER);
+		List<Pair<ItemStack,ItemStack>> items = Lists.newArrayList(
+				Pair.of(new ItemStack(Items.WOODEN_SWORD), repairWood),
+				Pair.of(new ItemStack(Items.WOODEN_PICKAXE), repairWood),
+				Pair.of(new ItemStack(Items.WOODEN_AXE), repairWood),
+				Pair.of(new ItemStack(Items.WOODEN_SHOVEL), repairWood),
```
mezz: I think this would be cleaner as a `Map`:
```java
Map<ItemStack, List<ItemStack>> items = new HashMap<>();
items.put(repairWood, Arrays.asList(
    new ItemStack(Items.WOODEN_SWORD),
    new ItemStack(Items.WOODEN_PICKAXE),
    ...
));
```

src/main/java/mezz/jei/plugins/vanilla/VanillaPlugin.java
```diff
@@ -133,17 +141,21 @@ public void register(IModRegistry registry) {
 		recipeTransferRegistry.addRecipeTransferHandler(ContainerFurnace.class, VanillaRecipeCategoryUid.SMELTING, 0, 1, 3, 36);
 		recipeTransferRegistry.addRecipeTransferHandler(ContainerFurnace.class, VanillaRecipeCategoryUid.FUEL, 1, 1, 3, 36);
 		recipeTransferRegistry.addRecipeTransferHandler(ContainerBrewingStand.class, VanillaRecipeCategoryUid.BREWING, 0, 4, 5, 36);
+		recipeTransferRegistry.addRecipeTransferHandler(ContainerRepair.class, VanillaRecipeCategoryUid.ANVIL, 0, 2, 3, 4 * 9);
```
mezz: `4 * 9` should be `36` like the other lines above it

src/main/java/mezz/jei/plugins/vanilla/anvil/AnvilRecipeWrapper.java
```diff
@@ -0,0 +1,77 @@
...
+
+	public AnvilRecipeWrapper(ItemStack leftInput, ItemStack rightInput, ItemStack output) {
+		this(leftInput, rightInput, output, -1);
+	}
+
+	public AnvilRecipeWrapper(ItemStack leftInput, ItemStack rightInput, ItemStack output, int levelsCost) {
+		this(leftInput, Collections.singletonList(rightInput), Collections.singletonList(output), levelsCost);
+	}
+
+	public AnvilRecipeWrapper(ItemStack leftInput, List<ItemStack> rightInputs, List<ItemStack> outputs) {
```
mezz: Similar to the API, this class should only need one constructor.

src/main/java/mezz/jei/plugins/vanilla/anvil/AnvilRecipeWrapper.java
```diff
@@ -0,0 +1,77 @@
...
+		this(leftInput, rightInputs, outputs, -1);
+	}
+
+	@SuppressWarnings("unchecked")
```
mezz: `@SuppressWarnings("unchecked")` should not be needed

mezz (changes requested):
Looks good. There are a few things I want changed before I test and pull it, but it's pretty much ready.

src/main/java/mezz/jei/api/IModRegistry.java
gigaherz: I think that makes the use of the API unnecessarily ugly, but your choice.
However, for context, the option to not have a cost shown wasn't about the recipe not having a cost, it was about the recipe's cost not being known beforehand, such as for enchanted item repair, where the cost depends on how many times the item has been repaired, and how many enchants it has.

src/main/java/mezz/jei/plugins/vanilla/anvil/AnvilRecipeMaker.java
gigaherz: I disagree, but it's your code so I'll fix it to your preference.

src/main/java/mezz/jei/plugins/vanilla/anvil/AnvilRecipeWrapper.java
gigaherz: If it isn't, I don't know how. I tried both `Lists.newArrayList` and `Arrays.asList`. So far as I can tell, Java is stupid and doesn't know how to handle generics in arrays, so it panics when it sees the varargs parameter type as `List<ItemStack>`.

mezz: I don't think it's possible to use `Arrays.asList` with genetics this way, you're right.
Try creating a new arraylist and then adding the inputs to it after creation.

src/main/java/mezz/jei/api/IModRegistry.java
mezz: Is it possible to automatically determine the cost, maybe using a fake `ContainerRepair`?

gigaherz:
I think this makes all the comments addressed.

mezz:
Is it possible to automatically determine the level cost, maybe using a fake `ContainerRepair`?

HenryLoenwind:
Yes, but that would be the cost for a fresh item, at best misleading ...

mezz:
It's better than no information, and means that nobody has to specify the level with the API.

Issue #202, opened 2016-03-07
String ID search filter


Prospector:
Not sure which is the last character used, but something like `&minecraft:red_flower` to list all of the blocks with the id of red_flower. Could also be generalised to `&red_flower` which could show (for example) `minecraft:red_flower` and `example:red_flower`

PR #806, opened 2017-04-11
Adds search filter for resource ids, Implements #202

way2muchnoise:
This will require a version bump as there are changes to the api (see `IIngredientHelper`).
It takes the whole resource id for now. Not sure if it should be that or just the resource path as there already is a mod id filter. (eg. `&minecraft` will now also give all blocks under the minecraft resource id and thus gives overlap with `@minecraft`).

It is set to disabled on default as I don't see a lot of people using it. The used prefix is `&` as proposed in #202 .

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,4 +5,4 @@ curse_project_id=238222
 
 version_major=4
 version_minor=3
-version_patch=1
+version_patch=2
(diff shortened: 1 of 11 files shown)


mezz:
Looks good. I think you should strip out the mod name so that there is no overlap with the mod/modID search.

mezz:
Looks good but needs some minor tweaks before it can be merged to avoid breaking addons.

Here's my protocol for supporting breaking API changes:
Bump the version number (to 4.3.2) and put a line at the end of the javadoc line for new method `@since JEI 4.3.2`.
Surround calls to the new method like this 
```java
try { 
    /* new method */
} catch (AbstractMethodError ignored) { // older ingredient helpers do not have this method
    /* sane default behavior */
}
```

You can see I do this for `IIngredientHelper#cheatIngredient` and `.IIngredientHelper#copyIngredient` right now.

Issue #343, opened 2016-07-04
Add support for Creative Inventory Action as an alternative to /give


exaskye:
Using [Creative Inventory Action](http://wiki.vg/Protocol#Creative_Inventory_Action) JEI could spawn in items for players in creative, even on vanilla servers, as players in Creative can summon any item at will from the server. There's a blacklist, but it's rather small.

PR #807, opened 2017-04-11
Implements #343

way2muchnoise:
Use Creative Inventory Action when in creative on servers that don't have JEI installed, as that doesn't require the player to be op. Works like pick block on vanilla, the item will go to the first free slot on the hotbar or the selected one (when not enchanted, see `InventoryPlayer#getBestHotbarSlot()`).

### DIFF
--- a/src/main/java/mezz/jei/util/CommandUtil.java
+++ b/src/main/java/mezz/jei/util/CommandUtil.java
@@ -13,6 +13,7 @@ import net.minecraft.util.text.ITextComponent;
 import net.minecraft.util.text.TextComponentString;
 import net.minecraft.util.text.TextComponentTranslation;
 import net.minecraft.util.text.TextFormatting;
+import net.minecraftforge.items.ItemHandlerHelper;
 import org.apache.commons.lang3.StringUtils;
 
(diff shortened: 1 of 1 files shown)


mezz:
Thanks!
Tested and this works similarly but not quite like pick block, there are some problems. If you grab a single item a few times, it will fill each slot instead of selecting an existing slot, and once the hotbar is full it always chooses the selected slot and replaces it without moving the old item into the inventory.
Is there a way to make it work more like the creative menu, where it sets the item held in your mouse?

mezz:
When I tested this time, it only cheats in items if you are in the "Survival Inventory" in creative. If you're looking at a creative page, nothing is cheated in. Also you can't cheat a stack, it's always 1 item.
I thought about it a bit and I would like this to work exactly like the `/give` mode if possible, the creative-inventory-action should be invisible to the player and work consistently with the rest of JEI's cleat modes.

mezz:
Looks good. 
I'm going to tweak it a bit so that JEI tries to use `/give` by default and uses this as a fallback, just to try to keep things consistent as possible.

PR #877, opened 2017-05-31
Added a safety check for BasicRecipeTransferInfo

sokratis12GR:
This prevents many unwanted crashes for bigger inventories, because values can exceed the slot limit and will throw an IndexOutOfBoundsException.

### DIFF
--- a/src/main/java/mezz/jei/transfer/BasicRecipeTransferInfo.java
+++ b/src/main/java/mezz/jei/transfer/BasicRecipeTransferInfo.java
@@ -53,8 +53,10 @@ public class BasicRecipeTransferInfo<C extends Container> implements IRecipeTran
 	public List<Slot> getInventorySlots(C container) {
 		List<Slot> slots = new ArrayList<Slot>();
 		for (int i = inventorySlotStart; i < inventorySlotStart + inventorySlotCount; i++) {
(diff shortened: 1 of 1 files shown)


mezz:
Thanks for the PR!
I'm not sure this is needed. Why not just give the correct inventory size?
Normally when I detect an error with using the API the goal is to crash as soon as possible to make it more obvious, I do not think it's good to cover up the issue.

sokratis12GR:
Actually you are right, it would be better to just check for the correct value, but it's a problem that when it errors it doesn't point exactly from which point of our code has the incorrect size. 
I.e I have 3-10x3-10 tables and the integration all worked fine until it reaches the 9x9 table, it errors that the index is out of bound while the correct values are still presented.

mezz:
Hmm ok. In that case I think you should detect if it is trying to go out of bounds and throw an exception that includes the container class name and how to fix the issue.

mezz:
Ideally I would probably want to test it on a real container right away, but I'm not sure every mod could pass in a valid container before the world is loaded.

sokratis12GR:
Eh, tbh I don't really think this pr is even needed, but yeah... should we close the pr ?

mezz:
Ok, I'll just close it then. Thanks

PR #921, opened 2017-07-05
Update ja_jp.lang

CrafterKina:
Should be "mistranslation" in en_us
```
key.jei.previousPage=Show Next Page
key.jei.nextPage=Show Previous Page
```


mezz:
Thanks! I will fix those in en_us

@way2muchnoise you wrote those :P

Issue #549, opened 2016-10-29
Feature Request:  Implement an API to accept an alternate item sorting comparitor.


IMarvinTPA:
I'm trying to figure this out myself, but I think I'm in over my head on it right now:
Add an API to allow an alternate sort system that can be used to override the default sort order in IngredientBaseListFactory.  May also need a method to force JEI to reload after sort order change.

The next trick will be getting NEI updated to feed its sorting system into this API to enable the sort panel to control the displayed sort order of items in the item panel.

Thanks,
IMarv

mezz:
Rather than interfacing with NEI, I think choosing the sort order should be something built into JEI. Adding new sorting methods could be part of the API.

IMarvinTPA:
Yes, that works too.
Maybe heavily borrow NEI's sorting API if the licenses are compatible?

Update:  I have item sorting implemented.  The config screen is ugly.  Now to figure out how to use the InvTweaks API and add it to the project...

IMarv

IMarvinTPA:
@mezz 
I'm not sure how to configure the project to not require adding Inventory Tweaks as a buildpath requirement.  But here is the updated code that I've gotten working.
Maybe it can be helpful to you.
[InventorySorting.zip](https://github.com/mezz/JustEnoughItems/files/560607/InventorySorting.zip)

The configuration screen for the sorting is ugly, but it functions.

Thanks,
IMarv

IMarvinTPA:
I put the code to do this in the linked zip in the previous comment.  I don't know how to do Git branching and push requests etc.  Some feedback would be appreciated if the above provided solution needs something else.

Thanks,
IMarv

PR #940, opened 2017-07-16
Add the ability to configure the sort order and to use inventory tweaks

IMarvinTPA:
Add the ability to configure the sort order and to use inventory tweaks's sorting.

https://github.com/mezz/JustEnoughItems/issues/549

### DIFF
--- a/src/main/java/mezz/jei/ingredients/IngredientListElementComparator.java
+++ b/src/main/java/mezz/jei/ingredients/IngredientListElementComparator.java
@@ -2,41 +2,281 @@ package mezz.jei.ingredients;
 
 import java.util.Comparator;
 
+import mezz.jei.Internal;
+import mezz.jei.api.ingredients.IIngredientHelper;
+import mezz.jei.config.Config;
 import mezz.jei.config.Constants;
 import mezz.jei.gui.ingredients.IIngredientListElement;
 import mezz.jei.util.Java6Util;
+import mezz.jei.util.Translator;
 import net.minecraft.item.ItemStack;
+import mezz.jei.startup.IModIdHelper;
+
+import mezz.jei.util.Log;
+import net.minecraft.item.Item;
+import net.minecraftforge.common.config.Property;
+import net.minecraftforge.fml.common.Loader;
+
+
+import java.util.*;
+
+import invtweaks.api.InvTweaksAPI;
 
 public final class IngredientListElementComparator implements Comparator<IIngredientListElement> {
 	public static final IngredientListElementComparator INSTANCE = new IngredientListElementComparator();
 
 	private IngredientListElementComparator() {
-
 	}
+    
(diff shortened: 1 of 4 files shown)


ChloeDawn:
Congrats, you added a hard dependency for a single "feature".

adrwa:
Why have you added a dep for inventory tweaks to JEI of all the things to dep?

IMarvinTPA:
So you can have the inventory panel on the right of the screen sort the items by how you have your inventory tweaks tree setup.  I'm working on a very large tree to sort basically everything.  It is nice having all of my iron/copper/whatever ingots grouped together.
I don't know enough to know how to make it a "soft" dependency if you mean for building it.  It runs just fine without Inventory Tweaks added to an actual run.

TehNut:
JEI has no dependencies, compile or otherwise, other than Forge and Minecraft. I highly doubt this will be merged if it adds one. 

This feature should be accomplished by exposing new methods to the API and having them added via plugins. *Not* in the core of the mod.

src/main/java/mezz/jei/ingredients/IngredientListElementComparator.java
```diff
@@ -2,41 +2,281 @@
...
+                String name2 = o2.getDisplayName();
+                return name1.compareTo(name2);
+            }
+        });
+		
+		//Inventory Tweaks needs to be part of the build environment for this bit to work.
+		boolean	hasInvTweaks = false;
+		try {
+			hasInvTweaks = Loader.isModLoaded("inventorytweaks"); 
+		} catch (Exception ex) {
```
adrwa, line 198: Why is this a try catch? Loader.isModLoaded doesn't throw an exception

adrwa (changes requested):

src/main/java/mezz/jei/ingredients/IngredientListElementComparator.java
IMarvinTPA: It does during the gradle build test phase.  I could not find a way to detect the test was running to not call isModLoaded.

IMarvinTPA:
I wouldn't mind having an API to connect to, but it is beyond my comfort level with Java to build one.  Feel free to use this as a basis for one and I'll make a tiny mod to connect the two APIs.

mezz:
Sorry but this feature is going to take a lot more work to do cleanly, which is why I haven't gotten around to it yet.

IMarvinTPA:
Thank you for looking at it.  Having the inventory panel sort is a big help in developing a tree file.  There is no chest big enough and easy enough to work with to hold one of everything and it is hard to even fill it anyway.  I don't know if AE has added sorting by Inventory Tweaks back in 1.10 or later either.  It was there in 1.7.  At least now it should be easier for me to maintain my personal JEI jar.

PR #942, opened 2017-07-18
Added seperate overlay draw call to IRecipeLayoutDrawable

brandon3055:
With the way IRecipeLayoutDrawable currently works recipe tool tips are drawn in the same render call as the main recipe. The issue with this is if you have more that one recipe in the screen or you are rendering other items after the recipe the recipe tool tips will render under those other gui elements.
This pull request does not affect the original draw code (So not a breaking change) But it adds 2 new draw calls. one to render the recipe background and ingredients. The other to render the recipe overlay. This allows you to render recipe tool tips in a seperate render pass.


mezz:
Fixed here https://github.com/mezz/JustEnoughItems/commit/e199e41e71c27b96b851263c3b43d4455f100b26

PR #1049, opened 2017-11-19
Updated Russian translation

raspopov:
Translated all untranslated strings.
Fixed some errors, readability and now closer to english original.


mezz:
Thank you!

PR #1054, opened 2017-11-25
Updated Russian translation

raspopov:
Added config.jei.advanced.giveMode and config.jei.advanced.giveMode.comment.


mezz:
Thank you!

PR #1065, opened 2017-12-04
Added Check for disabled brewing recipes

kindlich:
### What this PR does
It adds a check to the Potion finder to check if the recipe actually can be brewed before putting it into the recipes list. It also adds all removed recipes into a private static Set

### What is this good for?
If a mod changes how vanilla potions are brewed JEI would still show them as it uses the PotionHelper. With the additional call, removed recipes won't be registered


mezz:
Thanks! Looks good to me.

Issue #1067, opened 2017-12-09
[Help] Implementing IRecipeTransferHandler


codetaylor:
**My Goal**

I have a custom workbench that has a typical 3x3 crafting grid with an extra slot for a tool (non-stacking). When I shift-click on the recipe transfer button, I would like maxTransfer to apply to the 3x3 crafting grid despite the tool slot not being stackable. For example: 64 wood goes into the 3x3 and 1 axe goes into the tool slot. 

**What I've Tried**

First, I implemented `IRecipeTransferInfo`, which worked as expected when not shift-clicking. When shift-clicking, however, only one of each recipe item appeared in the 3x3 grid due to the limitation of the tool slot holding at most one item.

Next, I implemented `IRecipeTransferHandler`, thinking that I could use that to achieve the control over the transfer process that I needed. After walking through an implementation, specifically the `BasicRecipeTransferHandler`, I have come to the conclusion (now correct me if I'm wrong) that the `IRecipeTransferHandler` just prepares data to send to the server, and that the actual execution of the action takes place on the server, outside of the handler.

This bit seems to be limiting the transfer to one item each because of the non-stacking tool slot:

```java
int maxRemovedSets = maxTransfer ? 64 : 1;
for (Map.Entry<Integer, ItemStack> entry : slotMap.entrySet()) {
  ItemStack stack = entry.getValue();
  if (stack.isStackable()) {
    Integer craftNumber = entry.getKey();
    Integer slotNumber = craftingSlots.get(craftNumber);
    Slot craftSlot = container.getSlot(slotNumber);
    int maxStackSize = Math.min(craftSlot.getItemStackLimit(stack), stack.getMaxStackSize());
    maxRemovedSets = Math.min(maxRemovedSets, maxStackSize);
  } else {
    maxRemovedSets = 1;
  }
}
```

So my question is how can I solve this problem in an elegant way?

I thought of sending two packets from the `IRecipeTransferHandler`, each with different data, but since the server's packet handler clears the crafting grid, I doubt this would work. I also thought of implementing my own packet and packet handler, but I wanted to ask before I did the work.

Thanks for your time.

mezz:
Currently the `IRecipeTransferHandler` has some limitations, mainly because the implementation is complex and sort of badly written. It works well though, which makes it hard to change.
[...]

Issue #275, opened 2016-05-16
Put items in craft grid even if there not all there


Dijkstra1-zz:
I do like how JEI show when do not have all of the items you need, but sometimes that not all ways right with items with meta/NBT data, and have to one item manually would better ten fill all of grid, maybe shift click the plus button twice would do this, that way you still get see what is missing.

xChitose:
Experiencing the same little issue with an item that has multiple uses.

adamk33n3r:
I would even just like this to put the items that I DO have into the crafting square. Then I could go get the other items I need and come back to finish it.

Banderi:
Yeah, I agree. This was an especially good thing in NEI for crafting interfaces that keep your items in place like the TC crafting table; huge QoL improvement that avoided having to manually transfer all of the items or hold them all off in your inventory - particularly now that TC tables have an overworld stored items rendering, so you can easily see what you put in that table without having to open it. Even more unfortunately, since Mezz said he disliked it, we don't have any transparent overlay showing where the items in the recipe go, so you'd have to remember the recipe in order to manually transfer all the items (another thing that in fact was actually a good ting, albeit having it customizable or disable-able would have been better, but arguably far from being better off not having at all).
Please make this happen!

MentalMouse:
Yeah, this is an issue with ingredients like "any potion" and "any enchanted book", but especially with Blood Magic's orbs (which are used in crafting grids a lot)!

Phoupraw:
Same [here](https://github.com/mekanism/Mekanism/issues/7962).

PR #1068, opened 2017-12-09
Boolean to ignore set restriction when transferring recipe items.

codetaylor:
PR in response to [https://github.com/mezz/JustEnoughItems/issues/1067](https://github.com/mezz/JustEnoughItems/issues/1067)

### What

This PR introduces a default interface method in [IRecipeTransferInfo](https://github.com/mezz/JustEnoughItems/pull/1068/files/b8fcabe53c232cf3a6449e15263ec8625560201e#diff-5e703c41e725492759df7818691265b1) that returns a boolean and can be overridden by implementations that want to transfer as many items as possible into the crafting slots.

The [PacketRecipeTransfer](https://github.com/mezz/JustEnoughItems/pull/1068/files/b8fcabe53c232cf3a6449e15263ec8625560201e#diff-a7dd13e632d5333ff541cd0aa19c3799) has been altered to carry the boolean flag to the server.

The recipe transfer logic contained in [BasicRecipeTransferHandlerServer](https://github.com/mezz/JustEnoughItems/pull/1068/files/b8fcabe53c232cf3a6449e15263ec8625560201e#diff-d60e3f216533fe7a914066833895f1c9) has been slightly altered to branch based on this new flag. Now, instead of keeping track of how many items were removed as complete sets, the items removed are tracked in a map and all the existing logic still applies.

The [PlayerRecipeTransferHandler](https://github.com/mezz/JustEnoughItems/pull/1068/files/b8fcabe53c232cf3a6449e15263ec8625560201e#diff-ee589af180f346a253fcb54efd4db614) sends a packet with the new flag hardcoded to false.

### Why

This is useful for implementing a container that doesn't need to have a complete set of recipe items transferred and instead wants to transfer as many items as it can even if one slot has fewer items. See this issue: [https://github.com/mezz/JustEnoughItems/issues/1067](https://github.com/mezz/JustEnoughItems/issues/1067)

### DIFF
--- a/src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferInfo.java
+++ b/src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferInfo.java
@@ -38,4 +38,11 @@ public interface IRecipeTransferInfo<C extends Container> {
 	 * Return a list of slots that the transfer can use to get items for crafting, or place leftover items.
 	 */
 	List<Slot> getInventorySlots(C container);
+
+	/**
+	 * Return false if the recipe transfer should attempt to place as many items as possible for all slots, even if one slot has less.
+	 */
+	default boolean requireCompleteSets() {
+		return true;
+	}
 }
--- a/src/main/java/mezz/jei/network/packets/PacketRecipeTransfer.java
+++ b/src/main/java/mezz/jei/network/packets/PacketRecipeTransfer.java
@@ -16,14 +16,14 @@ public class PacketRecipeTransfer extends PacketJei {
 	public final List<Integer> craftingSlots;
 	public final List<Integer> inventorySlots;
 	private final boolean maxTransfer;
-	private final boolean ignoreTransferLimit;
+	private final boolean requireCompleteSets;
 
-	public PacketRecipeTransfer(Map<Integer, Integer> recipeMap, List<Integer> craftingSlots, List<Integer> inventorySlots, boolean maxTransfer, boolean ignoreTransferLimit) {
+	public PacketRecipeTransfer(Map<Integer, Integer> recipeMap, List<Integer> craftingSlots, List<Integer> inventorySlots, boolean maxTransfer, boolean requireCompleteSets) {
 		this.recipeMap = recipeMap;
 		this.craftingSlots = craftingSlots;
 		this.inventorySlots = inventorySlots;
 		this.maxTransfer = maxTransfer;
-		this.ignoreTransferLimit = ignoreTransferLimit;
+		this.requireCompleteSets = requireCompleteSets;
 	}
 
 	@Override
@@ -50,7 +50,7 @@ public class PacketRecipeTransfer extends PacketJei {
 		}
 
 		buf.writeBoolean(maxTransfer);
-		buf.writeBoolean(ignoreTransferLimit);
+		buf.writeBoolean(requireCompleteSets);
 	}
 
 	public static void readPacketData(PacketBuffer buf, EntityPlayer player) {
--- a/src/main/java/mezz/jei/network/packets/PacketRecipeTransfer.java
+++ b/src/main/java/mezz/jei/network/packets/PacketRecipeTransfer.java
@@ -16,12 +16,14 @@ public class PacketRecipeTransfer extends PacketJei {
 	public final List<Integer> craftingSlots;
 	public final List<Integer> inventorySlots;
 	private final boolean maxTransfer;
+	private final boolean ignoreTransferLimit;
 
-	public PacketRecipeTransfer(Map<Integer, Integer> recipeMap, List<Integer> craftingSlots, List<Integer> inventorySlots, boolean maxTransfer) {
+	public PacketRecipeTransfer(Map<Integer, Integer> recipeMap, List<Integer> craftingSlots, List<Integer> inventorySlots, boolean maxTransfer, boolean ignoreTransferLimit) {
 		this.recipeMap = recipeMap;
 		this.craftingSlots = craftingSlots;
 		this.inventorySlots = inventorySlots;
 		this.maxTransfer = maxTransfer;
+		this.ignoreTransferLimit = ignoreTransferLimit;
 	}
 
 	@Override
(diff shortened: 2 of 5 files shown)


src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferInfo.java
```diff
@@ -38,4 +38,11 @@
 	 * Return a list of slots that the transfer can use to get items for crafting, or place leftover items.
 	 */
 	List<Slot> getInventorySlots(C container);
+
+	/**
+	 * Return false if the recipe transfer should attempt to place as many items as possible for all slots, even if one slot has less.
+	 */
+	default boolean requireCompleteSets() {
```
codetaylor: I used a default interface method to introduce the flag so that the change wouldn't break any existing implementations.

src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferInfo.java
```diff
@@ -38,4 +38,11 @@
 	 * Return a list of slots that the transfer can use to get items for crafting, or place leftover items.
 	 */
 	List<Slot> getInventorySlots(C container);
+
```
codetaylor: I used a default interface method to introduce the flag so that the change wouldn't break any existing implementations.

src/main/java/mezz/jei/network/packets/PacketRecipeTransfer.java
```diff
@@ -16,12 +16,14 @@
 	public final List<Integer> craftingSlots;
```
codetaylor: The packet now carries the flag to the server.

src/main/java/mezz/jei/transfer/BasicRecipeTransferHandlerServer.java
```diff
@@ -53,10 +50,19 @@ public static void setItems(EntityPlayer player, Map<Integer, Integer> slotIdMap
 			return;
 		}
```
codetaylor: When removing items, either as sets or as individual items, the results are now tracked in a map. The values are copies of the item stacks with their counts each set to how many were successfully removed from the container.

codetaylor: I tried to leave as much of the original logic in place as possible.

mezz:
Hey, thanks for the PR. I haven't had a chance to review until today.
It looks like this adds a lot of extra code, which I wasn't really expecting. Is there any way to combine logic from `removeItemsFromInventory` and `removeSetsFromInventory` ?

codetaylor:
You're welcome. :) Yeah, I'm pretty sure combining the logic from those two methods could be done. I'll look into it soon.

codetaylor:
I have had a chance to look at this more thoroughly and have provided my notes below.

The situation where complete sets are required is a breadth-first operation. Each item in each required set of items is visited until the max number of sets is satisfied or, if an item can't be found, items in the incomplete set are rolled back and the operation is terminated.

The situation where complete sets aren't required is a depth-first operation. Each item in the required set of items is visited and an attempt is made to transfer one full stack of each.

I believe that combining the logic of the two situations could be done by visiting each item in the required set (breadth-first) and subsequently branching inside the loop for each situation, simply ignoring missing items on each iteration if not transferring as sets. I also believe that this would create a mess of the logic with nested conditionals and be less efficient than branching outside of each loop.

I propose a new method be created to replace `removeSetsFromInventory` and `removeItemsFromInventory`. It will be called, similarly, `removeItemsFromInventory` and will be a hybrid of the two existing methods. It will return a `Map<Integer, ItemStack>` that contains a map of all `ItemStack` to be transferred, complete with quantities, and indexed by the container slot they are to be transferred to.

Pseudo-code as follows:

```
if transferring as sets

  while true

    for each item in set, try to transfer one item

      if we can't find item to transfer or if stack to transfer is at max capacity
        roll back any items in incomplete set
        return map of items to transfer

else if we don't require full sets

  for each item in set, try to transfer up to one stack of item
  return map of items to transfer
```

The returned `Map<Integer, ItemStack>` can then be tested, and, if found to be emtpy, the entire transfer operation can be terminated.

This should also, by my understanding, eliminate the need to do a pre-calculation of the `maxRemovedSets` because, during the operation of gathering the stacks to transfer, availability and capacity is checked.

This will consolidate and clean up my initial implementation.

Please let me know what you think of this approach. I will await your review before starting the work.

mezz:
That sounds good to me.
I have been thinking about adding something where you click the [+] button even if it is grey so that it tries to transfer whatever is available (https://github.com/mezz/JustEnoughItems/issues/275). I think the new method would be able to support that as well.

codetaylor:
Okay, I've implemented this and it fails in a way that I didn't foresee. Treating the complete set transfer as breadth-first works, obviously, as that's how it was originally implemented. However, treating the 'as many items as possible' transfer operation as depth-first fails when a recipe contains more than one slot requiring the same item. For example, the slab recipe requires three planks in three different slots and, using my implementation, only the first slot is populated.

It must be implemented in the same way as the existing logic: breadth-first. This means that the new logic can be integrated into the existing logic. I originally thought that by keeping the two separate, it would increase the readability of the code and be more efficient, albeit on a micro level. I now believe that, since the two implementations must be very similar, keeping the two separate won't do all that much for readability and will introduce redundancy.

I plan to re-implement this and combine the logic.

mezz:
Sounds good, thanks for the update.
I'm not surprised you ran into some issues, the whole problem being solved here is very complicated.

codetaylor:
This implementation passes all cases that I could think to throw at it. Now I'm tired. :)

codetaylor:
I just realized that some lines use spaces and others use tabs to indent. I can fix this to match the project formatting if you like. I think the project uses tabs. I don't want to keep lumping commits into this PR if I can avoid it, but I'll happily fix that up.

mezz:
This project uses tabs to indent.
It'll take me a little while to test and merge this anyway, so no rush.

mezz:
I am going to accept this and then work on cleaning up this code in general. 
There are some more things I would like to handle, but in the current state it is too hard to work with.
Since I have some free time over the winter break it feels like a good time to work on it.
Thanks again!

PR #1105, opened 2018-01-07
Updated Russian translation

raspopov:
Added config.jei.advanced.maxColumns and config.jei.advanced.maxColumns.comment


mezz:
Thank you!

PR #1141, opened 2018-02-18
Added en_ud.lang

The-Fireplace:


mezz:
Hm thanks, but I kind of think this should be automatically generated somehow so that this doesn't need to be maintained. I don't really want to have to update this any time a string is added or changed.

The-Fireplace:
I made a script to autogenerate it. It's on my GitHub.

mezz:
Can you make a gradle plugin that generates it when you build the mod jar?

src/main/resources/assets/jei/lang/en_ud.lang
```diff
@@ -0,0 +1,116 @@
...
+key.jei.showUses=sǝs∩ ɯǝʇI ʍoɥS
+key.jei.recipeBack=ǝdᴉɔǝᴚ pǝʍǝᴉᴧ ʎꞁsnoᴉʌǝɹԀ ʍoɥS
+key.jei.toggleCheatMode=pǝꞁqɐuƎ ᵷuᴉʇɐǝɥƆ ǝꞁᵷᵷo⟘
+key.jei.previousPage=ǝᵷɐԀ snoᴉʌǝɹԀ ʍoɥS
+key.jei.nextPage=ǝᵷɐԀ ʇxǝN ʍoɥS
+
+config.jei.default=ʇꞁnɐɟǝᗡ
+config.jei.valid=pᴉꞁɐᴧ
+
+config.jei.title=ᵷᴉɟuoƆ ƎWⱯNᗡO%M
```
mezz, line 37: `%MODNAME` is a string that gets replaced in-code, it should not be made upside down.
I don't think there is an automatic way to detect replacement strings like this since it can be any format.

mezz (changes requested):

src/main/resources/assets/jei/lang/en_ud.lang
KnightMiner: Assuming this is made into a gradle plugin, you can add a special exceptions list when it runs in gradle

The-Fireplace:
Now that's an interesting idea. I didn't even think about making this a
gradle plugin. I'll look in to it.

src/main/resources/assets/jei/lang/en_ud.lang
liach: @KnightMiner Not yet... It is still an applet. https://github.com/The-Fireplace/MC-en-UD-Generator

PR #1241, opened 2018-05-22
fix(pt-BR) "item list" to "ingredient list"

ghost:
Replaced "item list" with "ingredient list", and a few improvements.


mezz:
Thank you!

PR #1260, opened 2018-06-11
1.10

beno31900:
If you can please add how long a machine took takes to cook something when you open a recipe and make it a setting you can turn on and off if you want.


mezz:
Thanks for the report!
It looks like you made Pull Request instead of an issue.
Please open a new issue here: https://github.com/mezz/JustEnoughItems/issues/new
I am not really sure what you are asking, so please try to explain there in some more detail.

PR #1329, opened 2018-08-13
Improved Suffix Tree

BloodWorkXGaming:


mezz:
For anyone interested:
We looked at this together on Discord. It reduces the recursive stack depth when adding new strings to the suffix tree, and reduces JEI load times a bit on large packs like ATM3.

PR #1374, opened 2018-10-01
Add a hook for getting the hovered ingredient in the recipe gui

### LINKED ISSUE #1357: API suggestion - a way to get the hovered item in recipe GUI
I was looking for a way to get the currently hovered item from the recipe GUI. Sadly, [the IRecipesGui interface](mezz/JustEnoughItems/blob/f5acb392ccdbc5ed8b6de6181f30bf0e42442957/src/api/java/mezz/jei/api/IRecipesGui.java), the only interface in the public API that the recipe GUI implements, does not have a way to get the currently hovered item like the [IIngredientListOverlay](https://github.com/mezz/JustEnoughItems/blob/f5acb392ccdbc5ed8b6de6181f30bf0e42442957/src/api/java/mezz/jei/api/IIngredientListOverlay.java#L19) does. 
This is definitely possible if I use JEI internals, but that might not work at all in a few versions.

*Use case*: I reimplemented Botania's Corporea request keybind (press a button when hovering over and item to request it from a nearby Corporea Index, see Vazkii/Botania#2668), but sadly this can't be used when in the recipe GUI. I was thinking about this when making that PR, but I only decided I really want it when I actually made use of this feature in a playthrough and couldn't request some crafting ingredients easily.

Hubry:
Resolves #1357. The hook is simple, just wrapping the existing internal method that takes position arguments to avoid copying the logic. I hope the version bump was appropriate, seems like the few PRs done in the past that extended the API did that, so...

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,4 +5,4 @@ curse_project_id=238222
 
 version_major=4
 version_minor=13
-version_patch=0
+version_patch=1
--- a/src/api/java/mezz/jei/api/IRecipesGui.java
+++ b/src/api/java/mezz/jei/api/IRecipesGui.java
@@ -1,8 +1,9 @@
 package mezz.jei.api;
 
-import java.util.List;
-
 import mezz.jei.api.recipe.IFocus;
+
+import javax.annotation.Nullable;
+import java.util.List;
 
 /**
  * JEI's gui for displaying recipes. Use this interface to open recipes.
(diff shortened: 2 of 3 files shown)


mezz:
Looks perfect, thanks for contributing!
Have you tested this with your mod? I have struggled a lot with mouse position in guis not being where I expect, so I'd like to see confirmation that it's working as expected.

Hubry:
All testing I did so far was using debug mode, but I'm gonna give it a try with the actual use I had in mind.

Hubry:
I tested it with my Botania dev enviroment on all GUI scales and it seems to be working perfectly.
("your mod" it is not really my mod, just contributing a bit :P )

mezz:
Cool, sounds good to me.

Issue #51, opened 2015-12-05
Make compass and clocks useless in item list and recipes


mezz:
Viewing the items instead of crafting them is a bit cheaty.

4rz0:
Ha!
I was always wondering about that in NEI ;D

mezz:
Talked to fry and this should be possible in 1.9.

Prospector:
Couldn't you simply replace the icon in JEI with a static compass icon?

mezz:
The issue is that there is no static compass icon, I think I would have to create it.
It would be a solution for the compass and clock, but I really want a more general solution.
Maybe I can just hack something together for these two for now.

Prospector:
Could you use the first frame of the compass/clock png?

mezz:
From what I could see, there's some hard-coded magic with those textures when they're loaded. Mojang doesn't handle it very cleanly. I have to try it out though.

Prospector:
I don't really think that would make a difference. There is a .png image for the two items, and just get the width of that image (by default 16x16), then get the top 16x16 (or whatever other width) used as the icon. Though this is an oversimplification, the theory would work I think.

Darkhax:
Wouldn't it be possible to try and trick the item, similarly to how the clock goes crazy when you are in the nether?

mezz:
This is working in 1.9

PR #1388, opened 2018-10-28
Fix McJtyMods/RFToolsDimensions#242: Pass a null player to getTooltip

josephcsible:
Similar to #51 where items leaked useful information in their render, some
items leak useful information in their tooltips too. Fix them the same way,
by passing a null player.

### DIFF
--- a/src/main/java/mezz/jei/plugins/vanilla/ingredients/ItemStackRenderer.java
+++ b/src/main/java/mezz/jei/plugins/vanilla/ingredients/ItemStackRenderer.java
@@ -13,7 +13,6 @@ import net.minecraft.client.gui.FontRenderer;
 import net.minecraft.client.renderer.GlStateManager;
 import net.minecraft.client.renderer.RenderHelper;
 import net.minecraft.client.util.ITooltipFlag;
-import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.EnumRarity;
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.text.TextFormatting;
@@ -34,10 +33,9 @@ public class ItemStackRenderer implements IIngredientRenderer<ItemStack> {
 
 	@Override
 	public List<String> getTooltip(Minecraft minecraft, ItemStack ingredient, ITooltipFlag tooltipFlag) {
-		EntityPlayer player = minecraft.player;
 		List<String> list;
 		try {
-			list = ingredient.getTooltip(player, tooltipFlag);
+			list = ingredient.getTooltip(null, tooltipFlag);
 		} catch (RuntimeException | LinkageError e) {
 			String itemStackInfo = ErrorUtil.getItemStackInfo(ingredient);
 			Log.get().error("Failed to get tooltip: {}", itemStackInfo, e);


mezz:
Thanks for the PR!

tomdodd4598:
So is there any way round this if we _want_ to have information shown on the tool-tips of the items in the list? Apparently this change has stopped NC's radiation level info being shown.

josephcsible:
@turbodiesel4598 Making that not show was exactly the point of the change. You shouldn't be able to use and get the benefit of items that you don't actually have. (Or am I missing something?)

tomdodd4598:
The NC radiation levels on the tool-tips of items just tell you how much radiation the item produces - it's certainly not something I want to hide from the player. [Here's](https://youtu.be/4KDPInVMPcE?t=40) what it looked like before this change.

josephcsible:
@turbodiesel4598 Ah, so that's an intrinsic property of the item, not based on the state of the world? In that case, it should definitely be fixable. I'll start work on a PR.

tomdodd4598:
Yeh, that's right - thanks!

josephcsible:
@turbodiesel4598 Okay, the issue was that in your tooltip method, you were checking if `player` was null and bailing out if it was, even though you didn't need or use `player` anywhere else. I removed that in turbodiesel4598/NuclearCraft#403.

squeek502:
This breaks tooltips in The Spice of Life, which requires the player context:
- https://github.com/squeek502/SpiceOfLife/issues/157
- https://github.com/Cazsius/Spice-of-Life-Carrot-Edition/issues/42

I guess TSoL could 'fix' this by using `Minecraft.thePlayer` instead of `event.getPlayerEntity()`, but that seems like piling a hack on top of a hack (i.e. it would break if someone wants to manually fire ItemTooltipEvent with a player that's not the local player [an edge case to be sure, though]; EDIT: this could also be mitigated by doing `player = event.getPlayerEntity() != null ? event.getPlayerEntity() : Minecraft.thePlayer`). 

Sending a null player seems like overkill for what is almost certainly a very special case. Surely there's a better way to signal 'this is coming from JEI' than discarding relevant context from the event that mods can and do expect to be included?

josephcsible:
@squeek502 The player parameter to getTooltip is explicitly marked as `@Nullable`, so if your code only works right when it's not null, that's 100% a bug on your end.

Anyway, when an item is in JEI, it doesn't belong to the player, so it doesn't make sense to call getTooltip as if it did. Why does your mod even need a player object to display the sorts of information that it does?

squeek502:
My mod uses the player field to lookup information specific to that player and display it (e.g. 'You have eaten this food X times'). The Spice of Life has existed since 1.6.4 and the player has always been supplied in NEI/JEI `ItemTooltipEvent` calls, so it seems weird to make the change in this direction, as it's likely to cause more problems than it solves due to changing something mods almost certainly expect not to change.

squeek502:
@josephcsible for RFTools, couldn't you do something like compare the tooltip event's ItemStack to JEI's `IIngredientListOverlay.getIngredientUnderMouse()` and not display the 'hidden' information if they are the same?

josephcsible:
@squeek502 Since you're displaying information that's relevant to the player looking at the item, regardless of who possesses the item, I think just unconditionally using `Minecraft.thePlayer` in your code is the right way to fix this.

To me, comparing with `IIngredientListOverlay.getIngredientUnderMouse()` seems like the hack-on-top-of-a-hack, and I'm not sure if it would even work.

squeek502:
Will wait to see what @mezz has to say.

mezz:
Unfortunately, in general I think this is causing more issues than it is solving, because very few mods keep sensitive data on the tooltip.
If mods want to be secretive about their tooltip data, it might be a better idea for them to have some checks to see if they are "real", rather than make everything else handle this case specially.

TehNut:
What if JEI passed a custom `ITooltipFlag` that mods who wish to hide data could check against?

josephcsible:
@mezz Isn't this way more consistent with the compass/clock fix though?

@TehNut If that works, I have no objection to it.

squeek502:
@mezz reminder about this.

mezz:
Thanks for the reminder, I have just reverted this PR.
For mods that want to protect information, I'm pretty sure we can come up with a way to manually check to make sure it's not in the creative menu or JEI.
Custom `ITooltipFlag` isn't a terrible idea but it would require a hard dependency on JEI.

josephcsible:
@mezz Any update on the alternative way to hide info in JEI tooltips?

mezz:
JEI and the recipe book collect tooltips at startup before the world exists, so you can probably add a check like
```java
if (Minecraft.getMinecraft().world == null) {
```

PR #1402, opened 2018-11-19
Fix #1202 by changing enchantments to ingredients

### LINKED ISSUE #1202: Modded Enchanted Books Shuffled / Incorrect
# Versions

Forge: 14.23.3.2655
JEI: 4.9.1.168
Full modlist: https://pastebin.com/7mM64ZDY

# The Issue

The video below best explains it. Enchanted books from other mods are getting scrambled around in JEI, making it impossible to figure out what enchantments are *actually* available and from which mods.

[Video](https://www.youtube.com/watch?v=iERSZdoA5-8)

naqaden:
Thanks to @HenryLoenwind for the [suggestion](https://github.com/mezz/JustEnoughItems/issues/1202#issuecomment-410721993).


src/main/java/mezz/jei/plugins/vanilla/ingredients/EnchantDataRenderer.java
```diff
@@ -0,0 +1,83 @@
...
+import net.minecraft.client.util.ITooltipFlag;
+import net.minecraft.enchantment.EnchantmentData;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.init.Items;
+import net.minecraft.item.EnumRarity;
+import net.minecraft.item.ItemEnchantedBook;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.text.TextFormatting;
+
+public class EnchantDataRenderer implements IIngredientRenderer<EnchantmentData> {
```
mezz: You should pass in an instance of `IIngredientRenderer<ItemStack>` to the constructor of this renderer, and use it to render the enchanted books.

mezz (changes requested):
Thanks for contributing!
This looks good but there are a couple small things to fix before I can merge it.
1. We will need a way to bridge the gap with recipes that use enchanted books right now. I have something hacked into recipe lookups for items that hold fluids, we can add another case to support old recipes that use enchanted books there. See `mezz.jei.recipes.RecipeRegistry#getRecipeWrappers` for the fluidstack special case handling.
2. JEI's existing enchanted book anvil recipes need to be updated

naqaden:
Thanks for that guidance. I don't yet see where to update anvil recipes, so right now each enchanted book shows anvil recipes for all enchanted books.

There's also a small issue where the previously blank enchanted book on the JEI items screen now renders with Protection I on it.

mezz:
I am working on adding oredict as an ingredient as well.
I think this PR will be a bit easier after that change is in, could take a while longer to get everything working right though.

HenryLoenwind:
oredict may be a bit trickier, as there are instances were we'd have lists of oredicts and itemstacks. Would a way to lookup all ingredients' recipes of a cycling slot at once be better? (This may need a way for teh recipe to tell you if it is cycling ingredient alternatives or recipe variants.)

mezz:
I've thought a lot about that question, and I was thinking oredict is the right way forward because of tags in 1.13.
With tags, it's even easier to create small groups of items for recipes.

That said, I'm not really convinced it's the right way either. I have implemented it in a branch so I can play around with it and see what corner cases I run into. So far there are a few pretty big ones where mods don't use oredict but still have large lists of items.

I am going to break off some pieces of my oredict branch that might benefit this PR. 
After that I will try the other approach, which is improving the way cycled ingredients are displayed and  navigated.

mezz:
This PR is pretty close to done, so I will update it and merge it in.

mezz:
I ended up changing this pretty dramatically.
Now the "enchant" ingredient is pretty much hidden from mods and the players, it will just act like the regular enchanted books.
I tested it in a few packs and situations where mods are added or removed, and it seems to work as intended now. Mods will not have to add support, it just fixes the ingredient list issue and everything worked out.

PR #1408, opened 2018-11-25
Bookmarks

HenryLoenwind:
Please review.

This works, but no piece of code is ever finished...

I added a dispatcher for the left side of the screen. At the moment it only knows the bookmark renderer and there's not yet a method to add others. That method, and a method to "steal" space from the area should probably be added before merging this.

I'm not sure about the "hint" text. Maybe instead show it in the bookmark area when there are no bookmarks? (In that case the bookmark area also should respond to the centered search field.)

It'd be nice to get an exclusion area for the recipe gui's side-tab. At the moment I have some hardcoded extra margin for this case.

The BookmarkOverlay is based on my own code, not IngredientListOverlay. I leave this to you to change in the future if you like.

Cheat mode also works with bookmarks, but ingredient drag&drop doesn't. I can add it if you don't want to rewrite the bookmark list as IngredientListOverlay soon.

Formatting is my default (Ender IO...), just have your IDE reformat those files I changed.


mezz:
Thanks for the contribution!

Some history for anyone who ends up on this page, Henry created this as a [separate mod](https://github.com/HenryLoenwind/JEIBookmarks) and asked me how to avoid using JEI internal classes so it could be more stable. This has been a highly-requested feature for a while, and one I wanted to add, so we agreed to put it into JEI directly that it can be a stable and permanent feature. 

I will be fixing up the style and tweaking it a bit before merging and maintaining it.

HenryLoenwind:
BTW, I chose "A" because "B" conflicts with Baubles, which is used quite often.

mezz:
Hmm okay. I like mnemonics for hotkeys so I'll have to think about it more.

mezz:
progress so far:
 * [x] reformat code
 * [x] remove on-screen "hint" text and add dedicated icon button to toggle the bookmarks list on and off. add hints to the button tooltip depending on the situation
 * [x] hide bookmarks navigation if there are no bookmarks
 * [x] fix tooltips to show mod name
 * [x] refactor to share the optimized rendering, etc of the item list
 * [x] find a good hotkey
 * [x] figure out what to do about sharing left-screen space (may be unnecessary with the toggle button)
 * [x] restrict the bookmark hotkey so you can only remove bookmarks when hovering things in the bookmark area
 * [x] test against real mod packs to see if any other ideas or issues come up
 * [x] merge!

HenryLoenwind:
> I like mnemonics for hotkeys 

That's why the text was "**A**dd/Remove Bookmark" ;)

> figure out what to do about sharing left-screen space (may be unnecessary with the toggle button)

I prepared 2 things: (a) The LeftAreaDispatcher allows other contents to be added and (b) its displayArea.y can easily be shifted by any amount. I think if those 2 ways are exposed in the API we're good.

>  restrict the bookmark hotkey so you can only remove bookmarks when hovering things in the bookmark area

I spent extra effort to allow it to remove bookmarks from anywhere. Undoing that should be easy, replace BookmarkList.contains()/remove() with code that checks for object identity instead. (objects in the bookmark list always are copies) (PS: add() still needs the old contains()!)

mezz:
Sounds good. I'll set it back to "A" for now until someone comes along with a better idea.

I'll take a look at `LeftAreaDispatcher ` and see if I can get it integrated well with the ItemZoom mod.

I made the change to only allow removal from the BookmarkList.
There is a very minor issues with the micro-infinities ingredient from EnderIO, I think it does not actually make a copy so the identity comparison doesn't work perfectly in that case. I think it's good enough though.

mezz:
To get left area into the API, I need to put `IShowsRecipeFocuses` and `IClickedIngredient` in the API too.
I think I will hold off on it, toggling the bookmark list on and off with the button is easy enough to do.

mezz:
Hmm it doesn't really play nice with FTBUtils.

mezz:
I'll merge it as-is now. 
We will get good feedback of the remaining minor issues as people try out the beta JEI.

HenryLoenwind:
that'd probably be easily fixed with an API method like `void registerTopReserveForLeftSide(int y)`. Or maybe a `GatherTopReserverEvent`.

mezz:
Yeah I'm working on a API for that, similar to the gui tab exclusion areas.

PR #1430, opened 2018-12-09
Jei too slow and dont let me quit a world

BlackHoleTON618:
when i want to quit a world i simply cant because it says shuttşng down server for hours and does nothing and JEİ is too slow when i want to search a item i have 87 pages im using forge  1.12.2 23.5.2768 and i musing modified direwolf20 modpack which has extra 50 mods than direwolf20's what should i do for searching fast and quitting a world because i dont want to force minecraft to end task


mezz:
Thanks for the report!
Please open this as an Issue, not a Pull Request, so I can keep track of it the right way.
It looks like everything is running a bit slow, not just JEI, but we can discuss more in a new Issue.
https://github.com/mezz/JustEnoughItems/issues/new

PR #1434, opened 2018-12-15
A few minor changes in the U.S. English translation

Greg-21:
**In this commit I did the following:**
- changed a few things in the U.S. English lang file, mainly for consistency purposes
- removed the unnecessary line in all lang files in which it existed.

### DIFF
--- a/src/main/resources/assets/jei/lang/en_us.lang
+++ b/src/main/resources/assets/jei/lang/en_us.lang
@@ -41,7 +41,7 @@ key.jei.focusSearch=Select Search Bar
 key.jei.showRecipe=Show Recipe
 key.jei.showUses=Show Uses
 key.jei.recipeBack=Show Previously Viewed Recipe
-key.jei.toggleCheatMode=Toggle Cheating Mode
+key.jei.toggleCheatMode=Toggle Cheat Mode
 key.jei.previousPage=Show Previous Page
 key.jei.nextPage=Show Next Page
 key.jei.bookmark=Add/Remove a Bookmarked Ingredient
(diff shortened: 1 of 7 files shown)


src/main/resources/assets/jei/lang/en_us.lang
```diff
@@ -94,11 +93,11 @@ config.jei.advanced.debugModeEnabled.comment=Only useful for JEI developers, add
 config.jei.advanced.centerSearchBarEnabled=Center Search Bar
 config.jei.advanced.centerSearchBarEnabled.comment=Move the JEI search bar to the bottom center of the screen.
 config.jei.advanced.modNameFormat=Mod Name Format
-config.jei.advanced.modNameFormat.comment=How the mod name should be formatted in the tooltip for JEI guis. Leave blank to disable.
+config.jei.advanced.modNameFormat.comment=How the mod name should be formatted in the tooltip for JEI GUIs. Leave blank to disable.
 config.jei.advanced.maxColumns=Max Width
 config.jei.advanced.maxColumns.comment=The maximum width of the ingredient list.
-config.jei.advanced.maxRecipeGuiHeight=Max Recipe Gui Height
-config.jei.advanced.maxRecipeGuiHeight.comment=The maximum height for the recipe gui.
+config.jei.advanced.maxRecipeGuiHeight=Max Height
```
mezz: max width and max height are for different parts of the UI, so the names should reflect that somehow.

mezz (changes requested):
Thanks, mostly looks good. We should fix the max width/height names before merging these changes.

src/main/resources/assets/jei/lang/en_us.lang
Greg-21: Initially I intended to enrich the first option's name to match the second one, but because I didn't know which name should I give to it, I decided to shorten the second option to match the first one.
Do you have any suggestion for a name for these two options? I'm open to suggestions.

Greg-21: Initially I intended to enrich the first option's name to match the second one, but because I didn't know which name should I give to it, I decided to shorten the second option to match the first one. I think they look OK after my change, but you are right, they could be confusing to someone.
Do you have any suggestion for a name for these two options? I'm open to suggestions.

Greg-21: Initially I intended to enrich the first option's name to match the second one, but because I didn't know which name should I give to it, I decided to shorten the second option to match the first one. I think they look OK after my change, but you are right, they could be confusing to someone.
Do you have any suggestions for names for these two options? I'm open to suggestions.

Greg-21: Initially I intended to enrich the first option's name to match the second one, but because I didn't know which name should I give to it, I decided to shorten the second option to match the first one. I think they look OK after my change, but you are right, they could be confusing to someone.
Do you have any suggestions for names for these two options? I'm open to them.

mezz: Maybe these?
`Max Overlay Width`
`Max Recipe GUI Height`

Greg-21: They sound OK, so since I have completely no idea for names, I'm going to use your suggestions.

mezz:
Thank you!

mezz (approved):

PR #1449, opened 2019-01-02
Update Polish translation

Greg-21:
In this commit I improved the translation, based on of the recently changed strings in the U.S. English lang file.


mezz:
Thank you!

PR #1504, opened 2019-03-02
Add Favorite Item or mark to show first.

Ze2F:
example press "f" on item to make it to away show first slot.
_Sorry I not good English._


mezz:
JEI has a "Bookmark" feature in the beta versions. You can press "A" over an item to bookmark it.

Next time please open an "issue" here and not a "pull request": https://github.com/mezz/JustEnoughItems/issues

PR #1533, opened 2019-03-30
Updated Polish translation

Greg-21:
In this commit I added and translated the new string in the `1.12` branch.

EDIT: I had a little confusion when I was translating the new string (`key.jei.toggleEditMode`). Don't you think just "Toggle Hide Mode" should fit here better, since this setting is related to "JEI **Hide** Ingredients Mode"? :)


mezz:
Thanks!
Hide/Edit mode is not named very well, but I haven't come up with something better.
I think as long as you are consistent in the translation it should be good.

Greg-21:
You're right. I noticed that word "edit" is used only in string names (probably a leftover from previous JEI versions, where "Hide Ingredients Mode" was named "Edit Mode"), so I translated the new string in such a way that it will be consistent with the new naming.

Greg-21:
Yeah, you're right. I noticed that word "edit" is used only in string names (probably a leftover from previous JEI versions, where "Hide Ingredients Mode" was named "Edit Mode"), so I translated the new string in such a way that it will be consistent with the new naming.

mezz:
Sounds good, thanks.

PR #1587, opened 2019-06-01
Updated Polish translation (branch 1.12)

Greg-21:
In this commit I changed two strings to better fit to the Polish language, in which we have two plural forms in this case, and removed a leftover from English translation.


mezz:
Thanks!

PR #1588, opened 2019-06-01
Fixed spelling mistake

Greg-21:
In this commit I fixed spelling mistake of "vice versa" in nearly all language files in the `1.13` branch (in other branches this issue still exists).


mezz:
Thanks for the fix.
This text is only used to debug, and it was copied from somewhere else.
I'm fine with fixing it though.

PR #1604, opened 2019-06-24
Updated Polish translation (branch 1.14)

Greg-21:
In this commit I added and translated the 3 new strings, and improved some sentences to increase their readability.

EDIT: I also changed two additional strings, one for similar reason as in #1586.


mezz:
Thank you!

PR #1609, opened 2019-06-26
Updated Polish translation (branch 1.13)

Greg-21:
In this commit I made the same changes as in #1604, with the exception of adding those 3 new strings, since they are redundant in JEI for Minecraft 1.13.


mezz:
Thank you!
The 1.13 will no longer be maintained now that 1.14 is available. It is not a long-term version to support.

Greg-21:
Good to know :) You just didn't change the main branch on GitHub from 1.13 to something else, so I was thinking that you still plan to support that version. And personally I don't like supporting old versions, so this was supposed to be the last translation improvement for 1.13.

PR #1610, opened 2019-06-26
Updated Polish translation (branch 1.12)

Greg-21:
In this commit I made the same changes as in #1604, with the exception of adding those 3 new strings, since they are redundant in JEI for Minecraft 1.12.


mezz:
Thank you!

PR #1653, opened 2019-07-29
Fix to get JEI running on 1.14.4

Siriuo:
Fix to get JEI running on 1.14.4 since the change to BrewingRecipe in forge. Not sure if this is a valid fix but it works :)


mezz:
Thanks for contributing!
This has been completed here: https://github.com/mezz/JustEnoughItems/commit/054ef7fb37ad3019f31947807142ba95e0457be4

PR #1660, opened 2019-08-07
Fix up for forge 1.14.4-28.0.45+(supercedes and includes #1653)

itsmeow:
This will work after the breaking changes of Forge 28.0.45 (events were moved) - hopefully this saves a few minutes, lol.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
 org.gradle.daemon=false
 
-mcversion=1.14.3
-forgeversion=27.0.17
+mcversion=1.14.4
+forgeversion=28.0.45
 forgegroup=net.minecraftforge
 mcp_mappings=20190630-1.14.3
 curse_project_id=238222
--- a/src/main/java/mezz/jei/gui/GuiEventHandler.java
+++ b/src/main/java/mezz/jei/gui/GuiEventHandler.java
@@ -4,14 +4,9 @@ import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
 
-import net.minecraftforge.fml.common.gameevent.TickEvent;
-import net.minecraftforge.client.event.GuiContainerEvent;
-import net.minecraftforge.client.event.GuiOpenEvent;
-import net.minecraftforge.client.event.GuiScreenEvent;
-import net.minecraftforge.eventbus.api.SubscribeEvent;
-import net.minecraft.client.Minecraft;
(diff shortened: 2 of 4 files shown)


mezz:
Thanks for contributing!
I don't think this solution is correct, it seems to be using the input ingredient twice and ignoring the potion input ingredient.
I have submitted a fix here: https://github.com/mezz/JustEnoughItems/commit/054ef7fb37ad3019f31947807142ba95e0457be4

itsmeow:
Alright thanks, (although it isn't my fix, that was from #1653). Thanks for the update.

itsmeow:
Hold on, this doesn't work on 28.0.45, which introduced breaking changes to Forge.

ProsperCraft:
We really need to make the hop over to .45 because of the forge breaking changes, several mods have updated there and only work with .45 and to continue testing we need to go there.

mezz:
I am unable to update to Forge .45 because of a ForgeGradle issue with that version of Forge in dev. I get an exception during this stage of setting up the project:
```
Running 'SideStripper'
java.lang.IllegalArgumentException
	at org.objectweb.asm.ClassVisitor.<init>(Unknown Source)
	at org.objectweb.asm.ClassVisitor.<init>(Unknown Source)
	at org.objectweb.asm.tree.ClassNode.<init>(ClassNode.java:164)
	at org.objectweb.asm.tree.ClassNode.<init>(ClassNode.java:150)
	at net.minecraftforge.gradle.mcp.function.SideAnnotationStripperFunction.execute(SideAnnotationStripperFunction.java:104)
```
When Lex or cpw fix that, I will be able to update.

itsmeow:
This PR is basically just an update to .45 aside from the Brewing stand thing, so without that it would function as a fix. I haven't seen that error, but I use eclipse so probably that's why. Thanks for all!

mezz:
Unfortunately I think this is caused by a ForgeGradle update a few days ago, and I cannot downgrade it because it's not really versioned. It should work for you as long as you don't update the Forge version or ForgeGradle.

mezz:
This also means my build system will not be able to build it, even if you or I submit a fix :/. Unfortunately we will just have to wait a little bit for the fix.

mezz:
I submitted an issue here: https://github.com/MinecraftForge/ForgeGradle/issues/611

PR #1671, opened 2019-08-11
Add deobf jar to artifact output

Vazkii:
This is required for any mod that wants to dep on JEI, as importing non-deobf jars will type error.


TehNut:
Shouldn't you be using `fg.deobf`? 

https://github.com/mezz/JustEnoughItems/wiki/Getting-Started#dependencies-using-fg3-minecraft-1132

Vazkii:
TIL that's a thing. My bad.

PR #1790, opened 2019-11-06
Correct the repair material for Elytra

### LINKED ISSUE #1773: [1.14] Elytra Anvil Repair Recipe is Incorrect
In playing in modded survival 1.14 on my public Geolosys Testing server, I was curious to see if Elytra were given any crafting uses by the few mods we have. Then I noticed that the repair item was still Leather from the pre-1.13 era. Just thought I'd through it out there that that's still an issue. Here is some photographic proof if it helps:
______
## Photos

What is shown:

javaw_muknr37CbB

My attempt at following this recipe:

javaw_t93UqamIeb

The actual recipe:

javaw_SoHPFB0IS0

_____

## Possible Source of the Problem

https://github.com/mezz/JustEnoughItems/blob/f520f423d9e518eed0a5a1067999b430d519fd70/src/main/java/mezz/jei/plugins/vanilla/anvil/AnvilRecipeMaker.java#L134

I'd be glad to make a PR with the new Phantom Membran if it makes your life easier!!

oitsjustjose:
This commit is in resolution of #1773 - in which Elytra were being repaired with leather. As of Minecraft 1.13 (in which the Phantom mob was added), Phantom Membranes are used to repair Elytra.

This commit simply creates a new repair item for Elytra, and registers it within AnvilRecipeMaker#getRepairRecipes

This commit can close #1773

### DIFF
--- a/src/main/java/mezz/jei/plugins/vanilla/anvil/AnvilRecipeMaker.java
+++ b/src/main/java/mezz/jei/plugins/vanilla/anvil/AnvilRecipeMaker.java
@@ -130,9 +130,9 @@ public final class AnvilRecipeMaker {
 			new ItemStack(Items.LEATHER_HELMET),
 			new ItemStack(Items.LEATHER_CHESTPLATE),
 			new ItemStack(Items.LEATHER_LEGGINGS),
-			new ItemStack(Items.LEATHER_BOOTS),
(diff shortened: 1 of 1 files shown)


mezz:
Thanks Jose!

oitsjustjose:
Gladly!

PR #1814, opened 2019-11-28
Initial memory usage optimization proposal

asiekierka:
This initial take on it reduces JEI's GeneralizedSuffixTree memory usage by 50-70% (depending on context); Enigmatica 2: Expert's in particular drops from 460MB to 165MB. The addition to load times is minimal - +30ms on a vanilla+JEI setup, +300-400ms on E2:E on my machine.

Personally, I would consider being even more "aggressive" in terms of the "edges" map size up to which we choose an ArrayMap over a HashMap; however, I am not sure how crucial performance is to that part of the code (an ArrayMap will always be significantly smaller RAM-wise, in this scenario, so perhaps always using it - plus an empty map singleton as used to not generate unneeded empty maps - would be the best approach), and the additional gains are only ~8MB in the "E2:E" scenario, so it might not be worth it in the end.

### DIFF
--- a/src/main/java/mezz/jei/suffixtree/Node.java
+++ b/src/main/java/mezz/jei/suffixtree/Node.java
@@ -17,11 +17,17 @@ package mezz.jei.suffixtree;
 
 import javax.annotation.Nullable;
 
+import it.unimi.dsi.fastutil.chars.Char2ObjectArrayMap;
 import it.unimi.dsi.fastutil.chars.Char2ObjectMap;
+import it.unimi.dsi.fastutil.chars.Char2ObjectMaps;
 import it.unimi.dsi.fastutil.chars.Char2ObjectOpenHashMap;
 import it.unimi.dsi.fastutil.ints.IntArrayList;
 import it.unimi.dsi.fastutil.ints.IntList;
 import it.unimi.dsi.fastutil.ints.IntSet;
+import it.unimi.dsi.fastutil.objects.ObjectCollection;
+
+import java.util.Collection;
+import java.util.Map;
 
 /**
  * Represents a node of the generalized suffix tree graph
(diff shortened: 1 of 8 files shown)


asiekierka:
Note: This only tackles the GeneralizedSuffixTree. Right now, in E2:E, JEI still occupies about 300MB of RAM idle total (down from about 600MB) - it is possible to reduce it further, to perhaps 250MB, but it requires some work (the Table and ListMultiMap in-memory storage form both seem to be far from optimal).

src/main/java/mezz/jei/suffixtree/Node.java
```diff
@@ -113,7 +119,13 @@ private boolean contains(int index) {
 	}
 
 	void addEdge(char ch, Edge e) {
-		edges.put(ch, e);
+		try {
+			edges.put(ch, e);
+		} catch (UnsupportedOperationException ex) {
+			// after trimToSize() call - fall back
+			edges = new Char2ObjectOpenHashMap<>(edges);
+			edges.put(ch, e);
```
mezz, line 129: Hitting this fallback a bunch could be pretty rough on performance, can you add a warning log here with the `ex` when it happens so we can trace down any mod that adds an ingredient super late?

mezz (changes requested):
Thanks asie! This looks like a very nice optimization with little downside.

src/main/java/mezz/jei/suffixtree/Node.java
asiekierka: Added! Sorry that it took me a while. I'm not sure about what the warning should say to the end user exactly, though. (Note that this will run once per changed Node, not once per mod addition or once per any entry addition.)

(Also, this fallback is the reason why I moved the memory optimization process to the very end of the JEI loading lifecycle)

PR #1827, opened 2019-12-17
Fix display name acquisition

pau101:
This PR corrects the usage of `ITextComponent#getUnformattedComponentText()` while `ITextComponent#getString()` (previously `ITextComponent#getUnformattedText()`) is the intended behavior. The former returns the contents of the specific text component not including siblings, leading to a potentially incomplete string.


mezz:
Sorry I missed this.

mezz:
Thanks!

PR #1830, opened 2019-12-23
Initial Port of JEI to 1.15.1

MoreThanHidden:
Ported JustEnoughItems over to 1.15.1 as I wanted to use it personally. 
(I would target a new branch but Github doesn't let me do that)

Only point of noting would be ItemStackFastRenderer is just using the vanilla equivalent of renderItemAndEffectIntoGUI otherwise just straight renames / equivalent functions.

Feel free to let me know if you want any changes or just reject if you have something already planned. 🤷🏻‍♂️

### DIFF
--- a/src/main/java/mezz/jei/render/ItemStackFastRenderer.java
+++ b/src/main/java/mezz/jei/render/ItemStackFastRenderer.java
@@ -2,21 +2,16 @@ package mezz.jei.render;
 
 import javax.annotation.Nullable;
 
-import net.minecraftforge.client.ForgeHooksClient;
-import com.mojang.blaze3d.platform.GlStateManager;
+import com.mojang.blaze3d.systems.RenderSystem;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.FontRenderer;
 import net.minecraft.client.renderer.ItemModelMesher;
 import net.minecraft.client.renderer.ItemRenderer;
 import net.minecraft.client.renderer.Rectangle2d;
 import net.minecraft.client.renderer.model.IBakedModel;
-import net.minecraft.client.renderer.model.ItemCameraTransforms;
-import net.minecraft.client.renderer.texture.AtlasTexture;
-import net.minecraft.client.renderer.texture.TextureManager;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.ResourceLocation;
-import net.minecraft.util.Util;
 
 import mezz.jei.config.IEditModeConfig;
 import mezz.jei.config.IWorldConfig;
@@ -49,7 +44,7 @@ public class ItemStackFastRenderer extends IngredientListElementRenderer<ItemSta
 	private void uncheckedRenderItemAndEffectIntoGUI(IEditModeConfig editModeConfig, IWorldConfig worldConfig) {
 		if (worldConfig.isEditModeEnabled()) {
 			renderEditMode(area, padding, editModeConfig);
-			GlStateManager.enableBlend();
+			RenderSystem.enableBlend();
 		}
 
 		ItemStack itemStack = element.getIngredient();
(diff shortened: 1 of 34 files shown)


TheLimePixel:
You're using JavaFX's points, that should be changed imo.

MoreThanHidden:
The vecmath library looks to have been removed and I think JavaFX is the better of the remaining implementations.
image
Otherwise could make a Point2D class ~~but why reinvent the wheel.~~

MoreThanHidden:
It got the better of me, as this is only used in this one location have created Point2D as a private static class.

tterrag1098:
Minecraft's Vec2f would work fine.

MoreThanHidden:
I did think of that, but will need to do the distance calculation separately and is a float (not that it matters).

src/main/java/mezz/jei/render/ItemStackFastRenderer.java
```diff
@@ -58,56 +51,9 @@ private void uncheckedRenderItemAndEffectIntoGUI(IEditModeConfig editModeConfig,
...
-			itemRenderer.renderModel(bakedModel, itemStack);
-
-			if (itemStack.hasEffect()) {
-				renderEffect(bakedModel);
-			}
-		}
-		GlStateManager.popMatrix();
-	}
-
-	protected void renderEffect(IBakedModel model) {
```
mezz, line 79: JEI optimizes rendering. It reduces the number of OpenGL calls by batching render stages (seen in `mezz.jei.render.IngredientListBatchRenderer#render`). This was done under the guidance of Fry.
Unless something changed in 1.15.1 (like Fry built these optimizations into vanilla), the code here is not exactly the same as `ItemRender.renderItemAndEffectIntoGUI` and this shouldn't be changed.

mezz (changes requested):
Thanks for contributing!

src/main/java/mezz/jei/render/ItemStackFastRenderer.java
MoreThanHidden: That's why I made note of it, there isn't a simple way to call a straight itemRenderer.renderModel in 1.15 (that I've seen) and my few attempts were less then optimal, but my thoughts were this had been implemented in vanilla but I can look into it further.

MoreThanHidden: Ah I see now, no that doesn't look to have been implemented in vanilla I've just pushed a change up that re-implements the same functionality in 1.15.

src/main/java/mezz/jei/render/ItemStackFastRenderer.java
```diff
@@ -49,65 +43,27 @@ private IBakedModel getBakedModel() {
...
 		{
-			GlStateManager.translatef(area.getX() + padding + 8.0f, area.getY() + padding + 8.0f, 150.0F);
-			GlStateManager.scalef(16F, -16F, 16F);
-			bakedModel = ForgeHooksClient.handleCameraTransforms(bakedModel, ItemCameraTransforms.TransformType.GUI, false);
-			GlStateManager.translatef(-0.5F, -0.5F, -0.5F);
-
+			RenderSystem.translatef(area.getX() + padding + 16.0F, area.getY() + padding, 150.0F);
+			RenderSystem.scalef(16F, -16F, 16F);
+			RenderSystem.translatef(-0.5F, -0.5F, -0.5F);
+			MatrixStack matrixStack = new MatrixStack();
```
pupnewfster: Not sure given not bothering to download the branch to experiment, but would it make more sense/be more performant to make the tralsations/scales directly to the matrix stack, rather then doing them via OpenGL?
```java
MatrixStack matrixStack = new MatrixStack();
matrixStack.func_227861_a_(area.getX() + padding + 16, area.getY() + padding, 150);
matrixStack.func_227862_a_(16, -16, 16);
matrixStack.func_227861_a_(-0.5, -0.5, -0.5);
```
(Note: func_227861_a_ which is the translate for a matrix stack, takes doubles instead of floats)

Also this way I believe it would have less GL calls, and you could remove the RenderSystem.push/pop matrix (I didn't bother adding the push/pop for matrix stack as it is a new matrix stack you are using that then gets GC'd after it is done being used). Also you can probably remove the `{`, `}` that are between the push/pop regardless.

src/main/java/mezz/jei/util/CommandUtilServer.java
```diff
@@ -113,7 +113,7 @@ public static void setHotbarSlot(ServerPlayerEntity sender, ItemStack itemStack,
 			}
 			ItemStack itemStackCopy = itemStack.copy();
 			sender.inventory.setInventorySlotContents(hotbarSlot, itemStack);
-			sender.world.playSound(null, sender.posX, sender.posY, sender.posZ, SoundEvents.ENTITY_ITEM_PICKUP, SoundCategory.PLAYERS, 0.2F, ((sender.getRNG().nextFloat() - sender.getRNG().nextFloat()) * 0.7F + 1.0F) * 2.0F);
+			sender.world.playSound(null, sender.serverPosX, sender.serverPosY, sender.serverPosZ, SoundEvents.ENTITY_ITEM_PICKUP, SoundCategory.PLAYERS, 0.2F, ((sender.getRNG().nextFloat() - sender.getRNG().nextFloat()) * 0.7F + 1.0F) * 2.0F);
```
pupnewfster: Not sure if `serverPosX` etc work properly here as I have not personally looked at what they do, so if they work fine like this they can probably be left as is; but if there is a slight difference in how they behave you may want to have this (and the call lower in this file), use the same call as before.
The getters for `posX`, `posY`, and `posZ` are as follows:
```
Entity#posX -> Entity#func_226277_ct_()
Entity#posY -> Entity#func_226278_cu_()
Entity#posZ -> Entity#func_226281_cx_()
```

src/main/java/mezz/jei/render/ItemStackFastRenderer.java
MoreThanHidden: Yeah I did test that and would probably improve it beyond what it was in previous versions, was just getting the same functionality and looking at how it was done in vanilla, plus trying to keep it readable, but I've changed it now.

pupnewfster: Ya I figured that it would work just wanted to make sure to mention it so that you would test it before replacing. Also figured you probably had not looked at translation stuff directly on the matrix stack due to it being a simple port, so figured I would mention it given I have lately been messing around with converting another mod from using GL calls in TESRs and the like to modifying the matrix stack instead.

MoreThanHidden: Thanks for the suggestion, I have been building up quite the dictionary of function names when porting some of my other TESRs as well, so will be ready to feed them in once we have a working MCP replacement. With 1.15 being new, going to take a bit of time to get used to new ways of optimizing OpenGL Calls.

src/main/java/mezz/jei/util/CommandUtilServer.java
```diff
@@ -157,7 +157,7 @@ private static void giveToInventory(PlayerEntity entityplayermp, ItemStack itemS
 				entityitem.makeFakeItem();
 			}
 
-			entityplayermp.world.playSound(null, entityplayermp.posX, entityplayermp.posY, entityplayermp.posZ, SoundEvents.ENTITY_ITEM_PICKUP, SoundCategory.PLAYERS, 0.2F, ((entityplayermp.getRNG().nextFloat() - entityplayermp.getRNG().nextFloat()) * 0.7F + 1.0F) * 2.0F);
+			entityplayermp.world.playSound(null, entityplayermp.serverPosX, entityplayermp.serverPosY, entityplayermp.serverPosZ, SoundEvents.ENTITY_ITEM_PICKUP, SoundCategory.PLAYERS, 0.2F, ((entityplayermp.getRNG().nextFloat() - entityplayermp.getRNG().nextFloat()) * 0.7F + 1.0F) * 2.0F);
```
pupnewfster: You missed switching to the getters here.

src/main/java/mezz/jei/util/CommandUtilServer.java
MoreThanHidden: Changed these to be on the safe side as serverPos isn't used elsewhere.

src/main/java/mezz/jei/util/CommandUtilServer.java
MoreThanHidden: Thanks yep my bad, fixed.

mezz:
Tested and looks good to me

isaac-smothers:
Trying to build this with `gradlew build` and getting the following:

```
* What went wrong:
Could not resolve all files for configuration ':apiCompileClasspath'.
> Could not find net.minecraftforge:forge:1.15.1-30.0.15_mapped_snapshot_20191105-1.15.1_at_6205343317139a3f64f2f07e1d668832419b5a70.
```

Not sure what to replace mcp_mappings with in gradle.properties

Issue #1838, opened 2020-01-06
[1.15.1] Shading problems with new Forge update


Dekamir:
Blocks are shaded wrongly with the new Forge update.
The following image should guide you.

https://i.hizliresim.com/Org120.png

Dekamir:
Image guide:
Left: 30.0.30 | Right: 30.0.24

ghost:
seems to be fixed with the latest Forge 30.0.41

PR #1842, opened 2020-01-10
Fix item shading for Forge 1.15.1-30.0.25+ #1838

MoreThanHidden:
Changed this back to using the Render System rather then using the MatrixStack.


MoreThanHidden:
Fixed in https://github.com/MinecraftForge/MinecraftForge/commit/1936eb1f368d48dbda1ccb53f46a21dd8f497a9d

PR #1950, opened 2020-04-30
add stonecutter support

hanetzer:
Not sure where to stick the 'click to open jei' button, and the gui is a bit
lackluster, but it works.
Signed-off-by: Marty E. Plummer <hanetzer@startmail.com>


mezz:
Thanks for contributing! The gui can be improved later if needed, this is good to have as a starting point.

PR #1963, opened 2020-05-15
1.12.2 fg3

RythenGlyth:
I'm trying to port jei 1.12.2 to ForgeGradle3 (the newest 1.12.2 builds use FG3). I think this involves
 - (removing `@SideOnly`)
 - changing the gradle file

I already changed the gradle file, but I ran into problems trying to compile: 
e.g. 
`The method initMissingImage() from the type TextureMap is not visible`  in https://github.com/mezz/JustEnoughItems/blob/1.12/src/main/java/mezz/jei/gui/textures/JeiTextureMap.java#L43
i got this error in the file 19 times
and many more

i could need some help, why this is happening and how i would fix it.
Thanks


RythenGlyth:
fixed the above named errors, but there are now rendering glitches

mezz:
Thanks for contributing!
I have a branch here, doing the same thing a little more cleanly:
https://github.com/mezz/JustEnoughItems/tree/1.12-FG3
Unfortunately I have run into the same issue as you, the resources don't get loaded.

mezz:
The resource issues has been solved here and will be merged into the main branch: https://github.com/mezz/JustEnoughItems/pull/2015
So I will close this PR

PR #1964, opened 2020-05-15
Silence "Recipe has no output." log for dynamic recipes 1.12.2

Davoleo:


Davoleo:
Asking @mezz for a review 👀

mezz:
@Davoleo can you give some context here? I'm not sure what this is solving.
If I had to guess, I'd assume there are some modded dynamic recipes that have handlers written for JEI. Is that right, and how would those be affected?

Davoleo:
Yeah, sure, sorry I didn't do this earlier
Basically when I'm using Forge json recipe factories in 1.12.2 to create some dynamic recipes that could have different outputs I don't want to set my output directly in the super-constructor of my recipe handler class, this is the reason why I set the first parameter to null
This however creates some problems with JEI since it logs errors because it tries to register a recipe that has a null output 

example snippet from a dynamic recipe handler class:
```java
public class ShapedMetalRecipe extends ShapedOreRecipe implements IRecipeMetal {

	private String resultType;

	public ShapedMetalRecipe(CraftingHelper.ShapedPrimer primer, String resultType)
	{
		super(null, ItemStack.EMPTY, primer);
		this.resultType = resultType;
		ModRecipes.shapedMetalRecipes.add(this);
	}

```

now, how would this fix this issue?
Well minecraft's `IRecipe` class contains a method called isDynamic which is by default set to false, and you can override it to set it to true
by adding that check  we make sure that whenever this method is overridden to true the recipe isn't loaded automatically and can be later integrated manually by the developer

Furthermore this is kind of a backport since you already do this kind of check in further versions like 1.14 and 1.15
https://github.com/mezz/JustEnoughItems/blob/1.15/src/main/java/mezz/jei/plugins/vanilla/crafting/VanillaRecipeValidator.java#L131-L133

mezz:
Thanks, that makes sense. I think my fear here is that since 1.12 is so far along, we'll run into issues with existing mods that depended on the old functionality.
Perhaps for now you can see if we can silence the error log when isDynamic is true, as a compromise for compatibility?

Davoleo:
mmh, do you mean that we should reverse the if condition and put it [here](https://github.com/mezz/JustEnoughItems/blob/8711de25c85cdfde3e71cc8b4566112774f0ad2a/src/main/java/mezz/jei/plugins/vanilla/crafting/CraftingRecipeChecker.java#L78)?
However doing it this way would mean that JEI tries to register recipes with null or empty output when isDynamic is true anyways, wouldn't it? 🤔 

The only other way to fix this without changing JEI source code would be to use a dummy itemstack inside the first super-constructor parameter that doesn't have any recipe (something like a command block), however doing this is a bit dangerous because it could create bugs in case someone tries to add a recipe to the Command Block

mezz:
I mean something like
```java
			if (recipeOutput == null || recipeOutput.isEmpty()) {
				if (!recipe.isDynamic()) {
					String recipeInfo = getInfo(recipe);
					Log.get().error("Recipe has no output. {}", recipeInfo);
				}
				return false;
			}
```

Davoleo:
oh right, I'll change it right now

mezz (approved):

PR #1967, opened 2020-05-18
[1.12.2] Added a config option to remove the recipe viewer

sor1n:
This should shave off a bit of memory in extreme cases, while still maintaining the item list and bookmark abilities. Essentially turning the mod into Too Many Items from long ago.

### DIFF
--- a/src/main/java/mezz/jei/config/Config.java
+++ b/src/main/java/mezz/jei/config/Config.java
@@ -1,7 +1,6 @@
 package mezz.jei.config;
 
-import javax.annotation.Nullable;
-import java.awt.Color;
+import java.awt.*;
 import java.io.File;
 import java.util.Collections;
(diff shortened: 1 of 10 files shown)


mezz:
Sorry but I don't think this is the best strategy to save memory.
I think most of the memory used by JEI is from the search index.
This adds some complexity and is a bit difficult to test that there are no holes in the logic where it would crash, especially relating to API calls and other things that are not easily tested by poking around.

sor1n:
Yeah, I just had to give my best guess to reduce the memory usage, since having too many mods JEI easily ends up taking so much memory it's insane. Thanks for commenting tho. You can close this if you wish.

mezz:
Thanks for taking a look into it yeah. I will close this one, but please feel free to try something else and take some memory measurements to confirm your guess.

PR #1971, opened 2020-05-30
Add Cosmetic Error Type for RecipeTransferHandler

Darkere:
This adds an option for custom recipe transfer handler to display an error while keeping the recipe transfer button enabled and usable. 

Example usage: https://gfycat.com/UnpleasantSomeDutchshepherddog

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,4 +8,4 @@ curse_project_id=238222
 
 version_major=6
 version_minor=0
-version_patch=0
+version_patch=2
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,4 +9,3 @@ curse_project_id=238222
 version_major=6
 version_minor=0
 version_patch=2
-
--- a/src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferError.java
+++ b/src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferError.java
@@ -26,9 +26,9 @@ public interface IRecipeTransferError {
 		USER_FACING,
 
 		/**
-		 * @since JEI version 6.0.2
 		 * Errors that still allow the usage of the recipe transfer button.
 		 * Hovering over the button will display the error, however the button is active and can be used.
+		 * @since JEI version 6.0.2
 		 */
 		COSMETIC
 
--- a/src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferError.java
+++ b/src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferError.java
@@ -26,6 +26,7 @@ public interface IRecipeTransferError {
 		USER_FACING,
 
 		/**
+		 * @since JEI version 6.0.2
 		 * Errors that still allow the usage of the recipe transfer button.
 		 * Hovering over the button will display the error, however the button is active and can be used.
 		 */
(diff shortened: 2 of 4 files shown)


Buuz135 (approved):

src/main/java/mezz/jei/gui/recipes/RecipeTransferButton.java
```diff
@@ -33,7 +33,7 @@ public void init(RecipeTransferManager recipeTransferManager, @Nullable Containe
 			this.recipeTransferError = RecipeTransferErrorInternal.INSTANCE;
 		}
 
-		if (this.recipeTransferError == null) {
+		if (this.recipeTransferError == null || this.recipeTransferError.getType().equals(IRecipeTransferError.Type.COSMETIC)) {
```
mezz: I'd like this to be pulled out into a helper function like `allowsTransfer(@nullable recipeTransferError)` that returns true or false. Then you can re-use it and make these `if` statements more readable.

Also, I think you can safely use `==` for the enum comparison here

mezz (changes requested):
Looks good to me and sounds like a reasonable, useful feature.
I have one small cleanup request before this can be merged.

src/main/java/mezz/jei/gui/recipes/RecipeTransferButton.java
Darkere: Good idea. I was struggling a lot with those ifs 😅

src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferError.java
```diff
@@ -23,7 +23,14 @@
 		 * Errors that the player can fix. Missing items, inventory full, etc.
 		 * Something informative will be shown to the player.
 		 */
-		USER_FACING
+		USER_FACING,
+
+		/**
+		 * Errors that still allow the usage of the recipe transfer button.
+		 * Hovering over the button will display the error, however the button is active and can be used.
+		 */
```
mezz, line 32: Please bump the JEI version to 6.0.2 and add a line to the docs here:
`* @since JEI version 6.0.2`
This will let users of the API know that they need to require this version if they use this feature.

src/main/java/mezz/jei/transfer/RecipeTransferUtil.java
```diff
@@ -61,6 +61,15 @@ private static IRecipeTransferError transferRecipe(RecipeTransferManager recipeT
 		return transferHandler.transferRecipe(container, recipeLayout, player, maxTransfer, doTransfer);
 	}
 
+	public static boolean allowTransfer(@Nullable IRecipeTransferError error){
```
mezz: Please rename this method to `allowsTransfer` and fix the spacing around braces for this method to match the code style.

mezz (changes requested):
Looks good! I have just a few small requests before we can merge this.

Darkere:
I think I managed to ping @ since 🤦 

Hopefully that's all the formatting. My IDE likes to format very differently :/

gradle.properties
```diff
@@ -1,11 +1,12 @@
...
+mcversion=1.15.2
+forgeversion=31.1.43
+forgegroup=net.minecraftforge
+mcp_mappings=20200411-1.15.1
+curse_project_id=238222
+
+version_major=6
+version_minor=0
+version_patch=2
+
```
mezz: For some reason Github says this whole file changed...?

mezz: I can clean this up myself and merge, one sec

Darkere: Ugh. Yeah, I didn't realize there was another JEI update today. And then I used GitHub to resolve the conflict instead of merging from the IDE. Didn't think that would be an issue.

mezz (approved):

PR #1983, opened 2020-06-20
Fix Cosmetic Error not showing

Darkere:
Followup to #1971

I've now finally fully understood why I struggled so much with that if statement (it didn't need to be changed at all…).

I've tested and confirmed it working this time around.


mezz:
Ah, right. No need to bump the version for this, I'll remove that and merge.

mezz (approved):

PR #1986, opened 2020-06-22
Update Polish translation

Greg-21:
In this commit I updated the Polish translation in the 1.12 branch. If you still support it, here is a comment about the changes.

Since word "zużycie" can also mean destroying something via using it, I don't like it in this case, so I changed it. Word "wykorzystanie" (like in the Windows Task Manager) would also suit here, but it is unnecessarily longer.


mezz:
Thanks!

PR #1995, opened 2020-07-03
Fix fluid amount formatting and anvil recipes

pupnewfster:
This PR fixes a few minor issues I have ran across in 1.16:
- Fix fluid translations not properly formatting the amount (as the translation text component only supports `%s` it seems)
- Simplified tag for item logic for items and added it for fluids
- Reworked the anvil recipe maker for vanilla gear to be based off the item's (or armor's) repair material to make it not require fixing between versions for existing items. (For example stone tools can now be repaired with black stone in vanilla)
- Fixed the anvil recipe maker printing errors when connecting to a dedicated server when an ingredient is empty (it is possible to make tags be empty, so if they are then it fails for things like shields)
- Added the repair recipes for netherite gear and armor, as well as, for turtle helmets.


mezz:
Thanks!

PR #2015, opened 2020-07-16
Add resources with Forge 1.12.2 and FG3

basdxz:
All thanks to hhyygg on the Forums
https://forums.minecraftforge.net/topic/36659-forge-not-finding-mcmodinfo-file/


mezz:
Thanks!
I tried this out and it works as expected.
I edited the comment to note that it's a hack... but I'll merge this into the main branch shortly.

PR #2032, opened 2020-08-09
Allow anvil recipes with no item in the right input slot

timschellenberg:
Talked to you, mezz, about this a week or so ago.
If the new collection null check in ErrorUtil is unnecessary for some reason I'll get rid of it, I figured that we still wanted to make sure that no null snuck in.

image


mezz (approved):

mezz:
Looks good to me, thanks for contributing!

PR #2041, opened 2020-08-13
Basic 1.16.2 Update

HyCraftHD:
Everything seems to work. Tooltips might be buggy, as the forge method of gui utils is currently not available


mezz (approved):

mezz:
Thanks for contributing!
Is there an issue I should watch to see when GuiUtils is complete?

HyCraftHD:
Probably not. I know that GuiUtils caches some item stack lookups, so if that gets implemented again it might be worth changing to it again

PR #2052, opened 2020-08-20
Updated texture of focused button state

Greg-21:
Minecraft 1.15 introduced a new indication of focused buttons, which is now a white outline, and since then the old bluish one that is still used by JEI clearly visually stands out from the rest.

Since no one has updated the JEI texture to make it match the vanilla one so far, I thought I could do it myself. If it is important, I should inform that I've used the OptiPNG 0.7.7 program to optimize the file.

This PR targets 1.16, but if you wish, I can create a PR for 1.15 as well.


mezz:
Thanks!

mezz:
Please create a PR for 1.15 as well so that you're credited with the fix.

Greg-21:
Done

Issue #2056, opened 2020-08-21
[1.16.2] JEI Crashes on New forge versions


BedrockLegends:
[crash-2020-08-21_11.09.02-client.txt](https://github.com/mezz/JustEnoughItems/files/5109929/crash-2020-08-21_11.09.02-client.txt)

When updating to forge 33.0.10, Minecraft will Launch fine and as soon as you try to join a world JEI crashes you to desktop

Forge version 33.0.10
JEI 1.16.2-7.1.0.11

cirogg:
Same here. something about Forge changed some method. Button.getHeight().

Ommina:
I'm guessing this https://github.com/MinecraftForge/MinecraftForge/commit/9118c3dd24f1eeb33b0dd957164d1905c2d001d2#diff-14883779050d9b45abc9fdd10ff7f7c0L16-L23 change to the Widget patch.  If there is some other means (rather than ATing the field) to get the height, I'm not immediately seeing it.

wanderingmage:
Same issue with Forge Version 33.0.19
Jei 1.16.2-7.1.0.12
It is pretty much the same as the initial claim. The crash log is identical.

mezz:
Thanks for the report! 
This has been fixed in the latest version.

PR #2057, opened 2020-08-21
Feature/fix not working on latest forge

JayJay1989:
This resolves issue #2056


mezz:
Thanks for contributing!
I actually fixed this just a little while ago when I updated the Forge version, so I'll close this PR.

Issue #2087, opened 2020-09-10
[1.16] Dark bucket items (DynamicBucketModel)


cech12:
The item models generated with the `DynamicBucketModel` looks darker in JEI in 1.16. 
![dark bucket](https://user-images.githubusercontent.com/16005054/92809406-e0415280-f3bc-11ea-85a4-0f446ee1ae8f.png)
All mods with buckets that use the `DynamicBucketModel` have this problem.

Here is also a repository of a test mod for this issue: https://github.com/cech12/BucketJEITest
And here is the used item model for the test item.
```
{
  "loader": "forge:bucket",
  "parent": "forge:item/default",
  "fluid": "minecraft:water",
  "textures": {
    "base": "minecraft:item/bucket",
    "fluid": "forge:item/mask/bucket_fluid"
  }
}
```

Tested in:
JEI version: jei-1.16.2-7.3.2.25
Minecraft: 1.16.2
Forge version: 33.0.59 

But it seems to be in other 1.16 versions, too.

cech12:
It looks like the items are rendered behind the dark background and some other custom models that extend the item layer seem to have similar behavior.

cech12:
![Unbenannt2](https://user-images.githubusercontent.com/16005054/92994802-ce1afd80-f4fd-11ea-86a5-6cb13b24e3db.png)

I disabled the background rendering temporarily and the bucket stays dark. So, it has nothing to do with the background rendering.

cech12:
Ok I found the problem. The `ItemMultiLayerBakedModel` in 1.16 returns true at `isGui3d()`. So, shading is activated for it (like for 3d models like blocks)
There are two possibilities to fix that:
1. (a bit hacky) fix in JEI where the new `ItemMultiLayerBakedModel` is excluded from the 3d rendering: https://github.com/mezz/JustEnoughItems/pull/2088
2. A change in Forge where the wrong return value of the `isGui3d` method is fixed.

cech12:
I fixed it now in a different way because using the `isGui3d` method was wrong. See the PR.
A change of Forges implementation should not be needed.

cech12:
merged in jei-1.16.2-7.3.2.28
thanks! :)

PR #2088, opened 2020-09-12
[1.16] fix shading bug of ItemMultiLayerBakedModel

cech12:
Bugfix for the issue https://github.com/mezz/JustEnoughItems/issues/2087
It is a bit hacky but it works for all classes that use the `ItemMultiLayerBakedModel`. It can be undone when Forge fixed the wrong return value of the `isGui3d()` method.


cech12:
My interpretation of the `isGui3d` method was wrong. It is only used for item entities. The method which is used to decide which shading should be used in the gui is called `func_230044_c_` (later mapping: `isSideLit`). I fixed it in the last commits. 
So, I think, there is nothing wrong in the Forge implementation. You can find the vanilla call of `func_230044_c_` in `net.minecraft.client.renderer.ItemRenderer::renderItemModelIntoGUI `

mezz:
Ok, that's pretty interesting. Thanks for the fix!

PR #2089, opened 2020-09-15
Add 1.16.3 compatibility

hanetzer:
mods.toml's version field was not being updated.
Mod loads on both 1.16.2 (forge 33.0.19) and 1.16.3 (forge 34.0.7).

Upper bound on minecraft version set to any 1.16.x version greater than
or equal to 1.16.2, may need finer tuning if a 1.16.4 version comes out
that breaks it.

Signed-off-by: Marty E. Plummer <hanetzer@startmail.com>


mezz:
Thanks!

PR #2109, opened 2020-10-06
Use GuiUtils.drawHoveringText to wrap long tooltips

pupnewfster:
This makes it so that tooltips on items can properly wrap again and any extra rendering people may do works. Additionally this PR bumps the min forge version to the one that reimplemented drawHoveringText


mezz:
Thanks!

PR #2133, opened 2020-10-24
[1.16] Optimize item stack rendering

Ocelot5836:
**This PR adds:**
- Batch rendering of item stacks using `IRenderTypeBuffer`

This change halves the time it takes to render item stacks in the overlay. Data is based on Minecraft's in-game frame time graph.

**Before:**
image

**After:**
image

### DIFF
--- a/src/main/java/mezz/jei/render/IngredientListBatchRenderer.java
+++ b/src/main/java/mezz/jei/render/IngredientListBatchRenderer.java
@@ -7,6 +7,7 @@ import java.util.List;
 
 import com.mojang.blaze3d.systems.RenderSystem;
 import net.minecraft.client.Minecraft;
+import net.minecraft.client.renderer.IRenderTypeBuffer;
 import net.minecraft.client.renderer.ItemModelMesher;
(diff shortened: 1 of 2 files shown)


src/main/java/mezz/jei/render/IngredientListBatchRenderer.java
```diff
@@ -156,6 +157,8 @@ public void render(Minecraft minecraft, MatrixStack matrixStack) {
 		TextureManager textureManager = minecraft.getTextureManager();
 		itemRenderer.zLevel += 50.0F;
 
+		IRenderTypeBuffer.Impl buffer = Minecraft.getInstance().getRenderTypeBuffers().getBufferSource();
```
pupnewfster: Given minecraft is one of the args of this method wouldn't it make more sense to use that instead of `Minecraft.getInstance()`

mezz:
Looks good to me, thanks!
This code is structured for batch rendering in 1.12 but I never got around to fully implementing it like this in later versions.

PR #2149, opened 2020-11-08
1.16.4

### LINKED ISSUE #2147: Version issue
So, the newest 1.16.4 version is for 1.16.3, said that the version need to be 34.0.3 or above and below 35.0.0, but there is no version of 1.16.4 are under 35.0.0

Raycoms:
Closes #2147

Adjust version range
Update correctly on change


mezz:
Thanks!

PR #2151, opened 2020-11-09
Implement event system ownership tracking (fix memory leak on reload)

### LINKED ISSUE #2148: [1.16.3] Huge memory leak on disconnect/reconnect
## Environment
**Minecraft version**: 1.16.3
**JEI version**: jei-1.16.4-7.6.0.52.jar

## Reproduction
1. Join a server
2. Note minimum used RAM after GC in F3 menu
3. Disconnect
4. Repeat

Every time you join, the minimum used RAM is increased with about 400MB, depending on how many items are defined in JEI. As you can imagine, this quickly eats up all the allocated RAM and starts severely stuttering, ultimately crashing with OOM.

I took a heap dump via JProfiler and noticed this:

image

Apparently, a new instance of `IngredientFilter` (and `RecipeManager`?) is allocated on each join. I tried to trace the incoming references in hopes I will find the source, but there's just too many. If you need any other information from the heap dump, please let me know and I will do my best to provide it.

ViRb3:
This PR reworks the JEI event system so that each event listener must have an associated owner. The ownership concept gives scope to the listeners, such that when an owner is unregistered, all listeners associated with it are also unregistered. This prevents all memory leaks caused by dangling (not unregistered) event listener subscriptions.

Practically, this change fixes at least #2148. With this PR applied, there is no longer a memory leak.

### DIFF
--- a/src/main/java/mezz/jei/JustEnoughItems.java
+++ b/src/main/java/mezz/jei/JustEnoughItems.java
@@ -33,7 +33,7 @@ public class JustEnoughItems {
 	private static void clientStart(IEventBus modEventBus, NetworkHandler networkHandler) {
 		JEIClientConfig.register();
 
-		EventBusHelper.addListener(modEventBus, ColorHandlerEvent.Block.class, setupEvent -> {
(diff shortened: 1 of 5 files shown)


mezz:
Thanks! We discussed on Discord and this looks good.

PR #2152, opened 2020-11-09
Improve network security

pau101:
This PR adds catch to all network exceptions to stop crash propagation, in addition to adding cheat permission check for item deletion.

### DIFF
--- a/src/main/java/mezz/jei/network/PacketHandler.java
+++ b/src/main/java/mezz/jei/network/PacketHandler.java
@@ -45,7 +45,7 @@ public class PacketHandler {
 			PacketIdServer packetId = PacketIdServer.VALUES[packetIdOrdinal];
 			IPacketJeiHandler packetHandler = serverHandlers.get(packetId);
 			packetHandler.readPacketData(packetBuffer, player);
-		} catch (RuntimeException e) {
+		} catch (Throwable e) {
 			LOGGER.error("Packet error for event: {}", event, e);
 		}
 		event.getSource().get().setPacketHandled(true);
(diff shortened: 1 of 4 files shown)


src/main/java/mezz/jei/network/PacketHandler.java
```diff
@@ -45,7 +45,7 @@ public void onPacket(NetworkEvent.ClientCustomPayloadEvent event) {
 			PacketIdServer packetId = PacketIdServer.VALUES[packetIdOrdinal];
 			IPacketJeiHandler packetHandler = serverHandlers.get(packetId);
 			packetHandler.readPacketData(packetBuffer, player);
-		} catch (RuntimeException e) {
+		} catch (Throwable e) {
```
mezz, line 48: I think catch `RuntimeException` is actually good here, we don't want to catch memory errors and other serious exceptions that could make debugging a crash very confusing.

src/main/java/mezz/jei/network/PacketHandlerClient.java
```diff
@@ -35,7 +35,7 @@ public void onPacket(NetworkEvent.ServerCustomPayloadEvent event) {
 			if (player != null) {
 				packetHandler.readPacketData(packetBuffer, player);
 			}
-		} catch (Exception e) {
+		} catch (Throwable e) {
```
mezz, line 38: I think catch `RuntimeException` is actually good here, we don't want to catch memory errors and other serious exceptions that could make debugging a crash very confusing.

mezz (changes requested):

mezz:
Discussed on Discord and this looks good, I don't like broadly catching exceptions but in this case it's how Vanilla and Forge do it. Communication between machines is sensitive and we should do what we can to prevent bad actors causing crashes directly in packet handlers.

mezz (approved):

PR #2153, opened 2020-11-10
Reduce memory used after exiting and loading a world

ViRb3:
`JeiReloadListener` is re-created and re-subscribed on each server join, causing a significant memory leak. It is not possible to unsubscribe it due to vanilla Minecraft limitations, but this PR works around the problem by keeping one reload listener and updating the state within it every time it is necessary.

Large objects such as `IngredientFilter` already occupy about 200MB of RAM in an average modpack. Due to the way Forge's event bus handles listener removal, 2 instances of these objects remain at all times, even though only one is used. This PR works around that, effectively saving half of that RAM space.

Reference: https://github.com/MinecraftForge/EventBus/issues/39


mezz (approved):
Looks good to me!

PR #2154, opened 2020-11-10
Fix recipe transfer exploit

pau101:
This PR adds a missing `Slot#canTakeStack()` condition to prevent a malicious client copying items out of ghost slots.


mezz:
Thanks!

PR #2157, opened 2020-11-11
Fix erroneous event listener removal

pau101:
This PR addresses the following:
1. Listener removal erroneously attempts to remove a `Consumer<T>` from a list of type `Subscription`
2. Owner removal depends on failed listener removal as otherwise a concurrent modification exception would be raised
3. A removal for an owner which is not registered results in a null deference
4. Listeners are only removed from the "FORGE" event bus regardless of what they are actually registered to


NielsPilgaard:
Say, did this issue happen to prevent servers from starting? I'm hunting a bug, apologies if I'm totally wrong here :P

ViRb3:
Not on my end, I've been hosting a server with JEI right before this PR without any issues.

PR #2191, opened 2020-12-30
Expose getUniqueIdentifierForStack in the IStackHelper API

pupnewfster:
Exposes `StackHelper#getUniqueIdentifierForStack` to the API via `IStackHelper#getUniqueIdentifierForStack` to allow for mods to do some more extensive caching while comparing two sets of items rather than recalculating the Unique ID for each pair of items.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,4 +10,4 @@ curse_project_id=238222
 
 version_major=7
 version_minor=6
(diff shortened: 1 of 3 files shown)


mezz (approved):
Looks good to me, thanks for the clean PR!

PR #2198, opened 2021-01-03
Deduplicate JEI plugins loaded from Jars containing multiple mods

covers1624:
Fairly simple. When there are multiple mods contained in a single Jar file, JEI will instantiate plugins contained within the jar multiple times.
Here we just collect to a Set in order to have a de-duplicated collection of plugins to load.

### DIFF
--- a/src/main/java/mezz/jei/util/AnnotatedInstanceUtil.java
+++ b/src/main/java/mezz/jei/util/AnnotatedInstanceUtil.java
@@ -1,8 +1,10 @@
 package mezz.jei.util;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 import net.minecraftforge.fml.ModList;
 import net.minecraftforge.forgespi.language.ModFileScanData;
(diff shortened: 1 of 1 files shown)


src/main/java/mezz/jei/util/AnnotatedInstanceUtil.java
```diff
@@ -28,7 +30,7 @@ private AnnotatedInstanceUtil() {
 	private static <T> List<T> getInstances(Class<?> annotationClass, Class<T> instanceClass) {
 		Type annotationType = Type.getType(annotationClass);
 		List<ModFileScanData> allScanData = ModList.get().getAllScanData();
-		List<String> pluginClassNames = new ArrayList<>();
+		Set<String> pluginClassNames = new HashSet<>();
```
mezz: This is going to ultimately change the order that plugins load, possibly exposing some issues in mod plugins that have undeclared dependencies on each other.
Can you use a `LinkedHashSet` here to ensure the order stays the same?

mezz (changes requested):
Thanks for the PR!

mezz (approved):

PR #2216, opened 2021-01-19
Fix a crash loop when closing the Recipe GUI

### LINKED ISSUE #2214: [1.16.4] Random crashes to desktop
A couple of players on my server are getting crashes to desktop whilst using JEI

Info:
MC 1.16.4
JEI 1.16.4-7.6.1.65
Forge: 35.1.36

Log
https://gist.github.com/Anteater46/e49ce18129e64e66b448d644525e0e53

Ocelot5836:
Fixes #2214.

### DIFF
--- a/src/main/java/mezz/jei/gui/recipes/RecipesGui.java
+++ b/src/main/java/mezz/jei/gui/recipes/RecipesGui.java
@@ -359,7 +359,7 @@ public class RecipesGui extends Screen implements IRecipesGui, IShowsRecipeFocus
 
 	private boolean handleKeybindings(InputMappings.Input input) {
 		if (KeyBindings.isInventoryCloseKey(input) || KeyBindings.isInventoryToggleKey(input)) {
-			onClose();
+			closeScreen();
 			return true;
 		} else if (KeyBindings.recipeBack.isActiveAndMatches(input)) {
 			back();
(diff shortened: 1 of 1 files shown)


mezz:
Thank you!
Can you please explain your fix?

Ocelot5836:
When Minecraft closes a screen, it calls `Screen#onClose`, which would then cause `Minecraft#displayGuiScreen` to be called if the parent screen was `null`.

*Example*
RecipesGui.java
```java
// If parent screen is null
if (parentScreen != null) {
	closing = true;
	minecraft.displayGuiScreen(parentScreen);
	parentScreen = null;
	closing = false;
} else {
	ClientPlayerEntity player = minecraft.player;	
	if (player != null) {
		// This calls Minecraft#displayGuiScreen(null) by proxy
		player.closeScreen();	
	}	
}
```
Minecraft.java
```java
Screen old = this.currentScreen;
net.minecraftforge.client.event.GuiOpenEvent event = new net.minecraftforge.client.event.GuiOpenEvent(guiScreenIn);
if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event)) return;

// Screen#onClose is called again
guiScreenIn = event.getGui();
if (old != null && guiScreenIn != old)
	old.onClose();
```
This kind of loop would only happen in very rare cases as seen in #2214

mezz:
Ok thanks for the explanation, I was able to follow your logic and it makes sense to me.

Issue #2165, opened 2020-11-16
JEI is not searching on Tags on a server 


MuteTiefling:
Thank you for reporting!

Version: jei-1.16.4-7.6.0.58.jar

$ doesn't seem to return any results when searching for tags on a server. Working fine in Single Player

https://github.com/NillerMedDild/Enigmatica6/issues/380

NielsPilgaard:
This is has been fixed by... something? No longer an issue in in-dev Enigmatica 6

MuteTiefling:
It seems random. Maybe related to something else, but it's started happening again on the public E6 server.

EndergamerHUN:
It also messes with color searching. Works in Single Player, not on servers.

Alwinfy:
Seems to be same underlying issue as this: https://github.com/mezz/JustEnoughItems/issues/2217#issuecomment-763172521

mezz:
Thanks for the report!
This has been fixed in the latest version of JEI by #2227

MuteTiefling:
Awesome, thanks for looking into this! :D

Xiarno:
Doesn't work for me and I'm using the latest server & client version possible. Tag are enabled in the settings.

PR #2227, opened 2021-01-30
Ensure tags are synced to client before JEI loads

### LINKED ISSUE #2177: Tags used befor it was bound during client load connecting on server
When a player join a server tags and recipes are sent with two separates packet. JEI starts when the recipes packet is received (using `RecipesUpdatedEvent`).
If the tag packaet hasen't been received yet and a mod plugin uses tags an exception (`java.lang.IllegalStateException: Tag <tag> used before it was bound`) will be thrown.

It may also be the cause of https://github.com/mezz/JustEnoughItems/issues/2165

Alwinfy:
Hopefully fixes #2177, #2165

### DIFF
--- a/src/main/java/mezz/jei/startup/ClientLifecycleHandler.java
+++ b/src/main/java/mezz/jei/startup/ClientLifecycleHandler.java
@@ -27,18 +27,13 @@ import mezz.jei.util.AnnotatedInstanceUtil;
 import mezz.jei.util.ErrorUtil;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.network.play.ClientPlayNetHandler;
-import net.minecraft.client.resources.ReloadListener;
 import net.minecraft.network.NetworkManager;
-import net.minecraft.resources.IReloadableResourceManager;
-import net.minecraft.resources.IResourceManager;
-import net.minecraft.world.World;
-import net.minecraftforge.client.event.ClientPlayerNetworkEvent;
+import net.minecraftforge.client.event.RecipesUpdatedEvent;
 import net.minecraftforge.event.AddReloadListenerEvent;
 import net.minecraftforge.event.TagsUpdatedEvent;
 import net.minecraftforge.event.world.WorldEvent;
 import net.minecraftforge.fml.loading.FMLPaths;
-import net.minecraftforge.resource.IResourceType;
-import net.minecraftforge.resource.ISelectiveResourceReloadListener;
+import net.minecraftforge.fml.network.NetworkHooks;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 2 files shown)


Vazkii:
Tested it on my server, can confirm it works.

src/main/java/mezz/jei/startup/ClientLifecycleHandler.java
```diff
@@ -89,43 +90,63 @@ public ClientLifecycleHandler(NetworkHandler networkHandler, Textures textures)
...
+				setupJEI();
+			}
+		});
+		// Listener for forge remote server
+		EventBusHelper.addListener(this, TagsUpdatedEvent.CustomTagTypes.class, event -> {
 			ClientPlayNetHandler connection = Minecraft.getInstance().getConnection();
-			if (connection != null) {
-				NetworkManager networkManager = connection.getNetworkManager();
-				worldConfig.syncWorldConfig(networkManager);
+			if (connection != null && !NetworkHooks.isVanillaConnection(connection.getNetworkManager()) && !Minecraft.getInstance().isIntegratedServerRunning()) {
```
mezz: To help clarify this logic, can you create a new enum and a helper function that returns `IntegratedVanillaServer`, `IntegratedForgeServer`, `RemoteVanillaServer`, or `RemoteForgeServer`? Then these conditions can just check using the enums.

mezz (changes requested):
Thanks for contributing, this looks great!
I have a small request to clean up some of the logic, but otherwise looks good to me.

mezz (approved):
super clean 👌

mezz:
@Vazkii can you test out this updated version? Then I can merge 👍

src/main/java/mezz/jei/startup/ClientLifecycleHandler.java
```diff
@@ -139,4 +147,27 @@ private void onRecipesLoaded() {
...
+
+		private ServerType(boolean isVanilla, boolean isIntegrated, Class<? extends Event> listenerClass) {
+			this.isVanilla = isVanilla;
+			this.isIntegrated = isIntegrated;
+			this.listenerClass = listenerClass;
+		}
+
+		public boolean connected() {
+			ClientPlayNetHandler connection = Minecraft.getInstance().getConnection();
+			boolean isVanilla = connection != null && NetworkHooks.isVanillaConnection(connection.getNetworkManager());
```
pupnewfster, line 168: It probably doesn't really matter because most people should be on 1.16.5 now and at the very least 1.16.4, but realistically this PR due to using this helper method should have bumped the min forge version from 1.16.3-34.0.3 to 1.16.3-34.0.18

Susanmsachs-commits:
@Alwinfy or @mezz how and where would I put in this code? I'm not tech savy but I really want to make JEI work for a server with my friends. Do I go in command prompt and copy the code; do I write it in notes; or do I configure an already made file of the server's code? Any advice or help would be greatly appreciated from anyone too.

Issue #2236, opened 2021-02-04
[1.16.4] Move Items button closes inventory. 


CrossVas:
Minecraft version: 1.16.4;
Forge version: 35.1.37;
JEI version: 1.16.4-7.6.1.70

![rec_1](https://user-images.githubusercontent.com/25727222/106944269-0c27f080-672f-11eb-9e3c-90e9d1ed2f5c.gif)

mezz:
Thanks for the report!

@Ocelot5836 I think this is related to your changes here: https://github.com/mezz/JustEnoughItems/pull/2216
Can you try to reproduce this issue?

mezz:
Fixed in version .71

PR #2237, opened 2021-02-04
Fix screen closing when trying to transfer recipe items

Ocelot5836:
This fixes the issue created by #2216 where the screen would close instead of returning to parent as shown by #2236. This PR fixes that by cancelling the close in the physical close method instead of the callback.


mezz:
Thanks for the quick fix!

PR #2238, opened 2021-02-08
Fix memory leak when reloading

malte0811:
Similar to #2153, see also #2148. I have verified that only one instance of `IngredientFilter` is present in heapdumps with this PR.


build.gradle
```diff
@@ -3,7 +3,6 @@ import se.bjurr.gitchangelog.plugin.gradle.GitChangelogTask
 buildscript {
 	repositories {
 		maven { url = "https://files.minecraftforge.net/maven" }
-		jcenter()
```
mezz: I don't think this should be changed

src/main/java/mezz/jei/ingredients/IngredientFilter.java
```diff
@@ -90,14 +91,17 @@ public IngredientFilter(
 			this.elementSearch.registerPrefix(prefixInfo);
 		}
 
+		WeakReference<IngredientFilter> weakThis = new WeakReference<>(this);
```
mezz: we can name it `weakIngredientFilter` and `ingredientFilter` instead of `weakThis` and `this2`

src/main/java/mezz/jei/ingredients/IngredientFilter.java
```diff
@@ -90,14 +91,17 @@ public IngredientFilter(
...
 		EventBusHelper.addListener(this, EditModeToggleEvent.class, editModeToggleEvent -> {
-			this.filterCached = null;
-			updateHidden();
+			IngredientFilter this2 = weakThis.get();
+			this2.filterCached = null;
```
mezz: we should check it is not null before calling anything on it

src/main/java/mezz/jei/ingredients/IngredientFilter.java
```diff
@@ -90,14 +91,17 @@ public IngredientFilter(
...
+			this2.updateHidden();
 		});
 
 		EventBusHelper.addListener(this, PlayerJoinedWorldEvent.class, playerJoinedWorldEvent -> {
-			this.filterCached = null;
-			updateHidden();
+			IngredientFilter this2 = weakThis.get();
+			this2.filterCached = null;
```
mezz: check it is not null first

mezz (changes requested):
thanks for contributing! I have come comments for changes before this can be merged

build.gradle
malte0811: I'll address the other comments in a bit, this change should stay though: [jcenter won't be available for much longer](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/), and since it isn't actually used I think removing it now is better than getting build failures when jcenter actually shuts down.

malte0811:
I found a few more cases of leaks through the event bus, with the last commit the active heap size after joining a world once and after joining it twice (with just JEI installed) is within 500 kB of each other, which I consider to be either an acceptable leak size or even measurement error.

One remark on indenting: I've tried to match the existing indent style per file. I assume the "correct" style would be tabs, but since `EventBusHelper` already uses 4 spaces everywhere I used that for my changes in that file too. The whole file should probably be changed to tabs, but IMO doing that in this PR is a bad idea just from a "readable diff" perspective.

mezz:
Sounds good. It's unfortunate we need to handle this so carefully to work around a bug.
I'll take a look now and hopefully merge this soon.

PR #2288, opened 2021-03-21
Update ForgeGradle & mappings

iTitus:
Update to ForgeGradle 4 and the newest MCP mappings.

Regarding the changes to `mods.toml`:

Current:
grafik

New:
grafik


iTitus:
When testing this you may need to re-import the gradle file. I had some problems with IntelliJ.

build.gradle
```diff
@@ -2,91 +2,99 @@ import se.bjurr.gitchangelog.plugin.gradle.GitChangelogTask
...
+	}
 }
 
 configurations {
-	apiCompile.extendsFrom(compile)
+	apiImplementation.extendsFrom(implementation)
+	apiRuntimeOnly.extendsFrom(runtimeOnly)
+}
+
+sourceCompatibility = targetCompatibility = compileJava.sourceCompatibility = compileJava.targetCompatibility = JavaVersion.VERSION_1_8
```
pupnewfster: For FG4 given it uses Gradle 6.8.+ this should be `java.toolchain.languageVersion = JavaLanguageVersion.of(8)`

build.gradle
```diff
@@ -2,91 +2,99 @@ import se.bjurr.gitchangelog.plugin.gradle.GitChangelogTask
...
-		version: "${project.mcversion}-${project.forgeversion}"
+			group  : "${project.forgegroup}",
+			name   : 'forge',
+			version: "${project.mcversion}-${project.forgeversion}"
 	])
-	implementation sourceSets.api.output
-	testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.2'
-	testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.6.2'
+	testImplementation platform('org.junit:junit-bom:5.7.1')
+	testImplementation 'org.junit.jupiter:junit-jupiter'
```
pupnewfster: This seems like an odd change

build.gradle
```diff
@@ -125,67 +133,68 @@ minecraft {
 	}
 }
 
-// prevent java 8's strict doclint for javadocs from failing builds
-if (JavaVersion.current().isJava8Compatible()) {
-	tasks.withType(Javadoc) {
+javadoc {
+	source = [sourceSets.main.allJava, sourceSets.api.allJava]
+	// prevent java 8's strict doclint for javadocs from failing builds
+	if (JavaVersion.current().isJava8Compatible()) {
```
pupnewfster: Someone should double check but I am pretty sure this check shouldn't be needed if the above comment about setting the java toolchain is done. (As in the if statement itself, the body of the if statement is probably still needed)

pupnewfster (commented):
Left a few comments about the changes to the build.gradle. One other thing that may be worth evaluating for mezz before merging this is whether or not he wants to while having the mappings get updated follow forge's lead and switch to mojang mappings from crowdsourced.

build.gradle
iTitus: you're right about this one.

build.gradle
iTitus: There are two possibilities and JUnit suggests both...
- https://junit.org/junit5/docs/current/user-guide/#running-tests-build-gradle-engines-configure
- https://github.com/junit-team/junit5-samples/blob/r5.7.1/junit5-jupiter-starter-gradle-groovy/build.gradle

build.gradle
iTitus: The toolchains are annoying because they override your set JDK in the IDE and download new JDKs into your (already huge) .gradle folder.
I don't need to download Java 8 if I have it installed already, thanks for nothing Gradle :)

iTitus:
If requested I can also do the `official` mappings.

actuallyasmartname (changes requested):
The git changelog gradle plugin's version has changed to 1.66 lmao
By the way, removing JCenter is probably good too since gradle builds perfectly without it

actuallyasmartname (approved):

build.gradle
mezz: I prefer keeping the version on both

build.gradle
```diff
@@ -125,67 +132,66 @@ minecraft {
...
-				"Implementation-Version": "${project.version}",
-				"Implementation-Vendor" :"mezz",
-				"Implementation-Timestamp": new Date().format("yyyy-MM-dd'T'HH:mm:ssZ"),
+				'Specification-Title'     : 'Just Enough Items',
+				'Specification-Vendor'    : 'mezz',
+				'Specification-Version'   : "${semver}",
+				'Implementation-Title'    : project.name,
+				'Implementation-Version'  : "${project.version}",
+				'Implementation-Vendor'   : 'mezz',
+				'Implementation-Timestamp': new Date().format('yyyy-MM-dd\'T\'HH:mm:ssZ'),
```
mezz: this one actually seems like an appropriate use for `"` unless they've been completely disallowed

mezz:
Looks good, thanks!
I still need to try this out before merging. I never trust what weird stuff gradle will do in reality vs in theory.

Let's not update the mappings to `official` in the same PR. I haven't really evaluated what it would look like yet, and it seems like a separate concern from the rest of this.

PR #2290, opened 2021-03-22
Add in support for showing shulker box coloring recipes

pupnewfster:
One thing to note, is that similar to how the existing tipped arrow recipe compat is, the shulker box recipe adder makes no attempt to validate that the special recipe is in fact registered as I couldn't really come up with a good/clean way to do so


actuallyasmartname (approved):
Nice job! @mezz should merge this anytime soon.

PR #2295, opened 2021-03-26
Add proper element validator to IngredientSortStages and SearchColors config entries

pupnewfster:
I made two minor changes to the client config:
1. I changed the default list for `SearchColors` from using `Arrays#asList` to `Lists#newArrayList` as I recall in the past night config having some issues with the list implementation in `Arrays#asList`
2. I added proper validators to `IngredientSortStages` and `SearchColors` so that if an element is not valid then it properly is handled/logged using forge's config system. Strictly speaking this means that the extra validation JEI does afterwards when using these lists is not needed

### DIFF
--- a/src/main/java/mezz/jei/config/ClientConfig.java
+++ b/src/main/java/mezz/jei/config/ClientConfig.java
@@ -4,6 +4,7 @@ import com.feed_the_beast.mods.ftbguilibrary.config.ConfigGroup;
 import com.feed_the_beast.mods.ftbguilibrary.config.NameMap;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
 import mezz.jei.Internal;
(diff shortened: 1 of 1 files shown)


src/main/java/mezz/jei/config/ClientConfig.java
```diff
@@ -202,4 +220,20 @@ private void syncSearchColorsConfig() {
 		final ColorNamer colorNamer = new ColorNamer(searchColorsMapBuilder.build());
 		Internal.setColorNamer(colorNamer);
 	}
+
+	private static Predicate<Object> validEnumElement(Class<? extends Enum<?>> enumClazz) {
```
mezz: I'd prefer to call it `enumClass`, the zz thing gives me bad flashbacks to java textbooks that try giving too-short names to everything.

mezz (changes requested):
Looks good to me, just a minor nit to fix before merging

PR #2298, opened 2021-04-03
Close the recipe screen even if parent screen is null

deirn:
When you open the recipe screen when still in world, `this.parentScreen` is null, making you stuck on the recipe screen with no way out.


mezz:
Thanks for contributing! Can you please explain the issue in more detail? I am not sure how to reproduce the bug this is fixing.

deirn:
So I want to make a thing that shows the recipe/usage for the block you look at. I called `IRecipeManager#createFocus` while the player in game, making `Minecraft#currentScreen == null`  which makes `parentScreen == null`.

mezz:
Thanks, that makes sense to me. 
We used to close the player's screen in this case, but it caused a loop bug reported here: https://github.com/mezz/JustEnoughItems/issues/2214
I think just setting the GUI to `null` should be safe here though, I'll check to see if it has any side effects.

mezz:
Looks like it should be fine, I'll merge this as-is.

PR #2306, opened 2021-04-11
Transition various string based APIs over to text components

pupnewfster:
As all these changes relate to transitioning strings to text components I put them in one PR though three separate commits to make it easier to see what changed what. If it is preferred I split this into three different PRs I can do so, though the version number will need to be bumped three times then instead of just once.

1. Add support for specifying ingredient info using text components. To get this working properly I also had to make the following changes:
- Fixed existing code not supporting text components with styles
- Fix variable replacement in translation text components only recognizing the contents before the first replacement.
2. Add support to IRecipeTransferHandlerHelper for using text components as the error message.
3. Switch IRecipeCategory to being able to return a text component for the title.


src/api/java/mezz/jei/api/recipe/category/IRecipeCategory.java
```diff
@@ -36,9 +37,22 @@
 	/**
 	 * Returns the localized name for this recipe type.
 	 * Drawn at the top of the recipe GUI pages for this category.
+	 * @deprecated Prefer calling and using {@link #getTitleAsTextComponent()}.
 	 */
+	@Deprecated
 	String getTitle();
```
pupnewfster, line 43: Should this be defaulted to:
```java
return getTitleAsTextComponent().getString();
```
I wasn't sure how clear it would be if it was that one of the get titles needs to be implemented, which is why I didn't do so for now.

src/api/java/mezz/jei/api/recipe/category/IRecipeCategory.java
```diff
@@ -36,9 +37,22 @@
...
 
+	/**
+	 * Returns a text component representing the name of this recipe type.
+	 * Drawn at the top of the recipe GUI pages for this category.
+	 * @since JEI 7.6.4
+	 */
+	default ITextComponent getTitleAsTextComponent() {
+		//TODO: When getTitle is removed this should just be rename to getTitle and not be defaulted anymore.
+		String title = getTitle();
+		return title == null ? null : new StringTextComponent(title);
```
pupnewfster, line 53: The recipe category tab has a null check on getTitle so I put one here as well, though I am not sure if it should be here as well or not.

src/main/java/mezz/jei/transfer/RecipeTransferErrorTooltip.java
```diff
@@ -17,9 +17,9 @@
 public class RecipeTransferErrorTooltip implements IRecipeTransferError {
 	private final List<ITextComponent> message = new ArrayList<>();
 
-	public RecipeTransferErrorTooltip(String message) {
+	public RecipeTransferErrorTooltip(ITextComponent message) {
 		this.message.add(new TranslationTextComponent("jei.tooltip.transfer"));
-		StringTextComponent messageTextComponent = new StringTextComponent(message);
+		IFormattableTextComponent messageTextComponent = message.deepCopy();
```
pupnewfster, line 23: This does a deep copy here because there is a chance a text component that is not supposed to be formattable is passed here. If you would rather I can make the transfer error helpers take formattable text components and specify in the JavaDoc that they are supposed to be allowed to be formatted.

mezz:
I did some code auto-formatting on the master branch, so I updated this PR to resolve conflicts

src/api/java/mezz/jei/api/recipe/category/IRecipeCategory.java
```diff
@@ -36,9 +37,22 @@
 	/**
 	 * Returns the localized name for this recipe type.
 	 * Drawn at the top of the recipe GUI pages for this category.
+	 * @deprecated Prefer calling and using {@link #getTitleAsTextComponent()}.
```
mezz: I like putting a version on deprecation warnings so people know how old it is
```suggestion
	 * @deprecated since JEI 7.6.4, prefer calling and using {@link #getTitleAsTextComponent()}.
```

src/api/java/mezz/jei/api/recipe/category/IRecipeCategory.java
mezz: What you have now is the right way, we shouldn't give this a default. Unfortunately new methods like `getTitleAsTextComponent` will always need a default implementation for compatibility, and it is confusing if they are both defaulted and point at each other.

src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferHandlerHelper.java
```diff
@@ -21,15 +22,37 @@
 	 * Create an error with type {@link Type#USER_FACING} that shows a tooltip.
 	 *
 	 * @param tooltipMessage the message to show on the tooltip for the recipe transfer button.
+	 * @deprecated Use {@link #createUserErrorWithTooltip(ITextComponent)} instead.
```
mezz:
```suggestion
	 * @deprecated Since JEI 7.6.4, use {@link #createUserErrorWithTooltip(ITextComponent)} instead.
```

src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferHandlerHelper.java
```diff
@@ -21,15 +22,37 @@
...
+	 */
+	IRecipeTransferError createUserErrorWithTooltip(ITextComponent tooltipMessage);
+
 	/**
 	 * Create an error with type {@link Type#USER_FACING} that shows a tooltip and highlights missing item slots.
 	 *
 	 * @param tooltipMessage   the message to show on the tooltip for the recipe transfer button.
 	 * @param missingItemSlots the slot indexes for items that are missing. Must not be empty.
 	 *                         Slots are indexed according to {@link IGuiItemStackGroup#getGuiIngredients()}.
+	 * @deprecated Use {@link #createUserErrorForSlots(ITextComponent, Collection)} instead.
```
mezz:
```suggestion
	 * @deprecated Since JEI 7.6.4, use {@link #createUserErrorForSlots(ITextComponent, Collection)} instead.
```

src/api/java/mezz/jei/api/registration/IRecipeRegistration.java
```diff
@@ -32,9 +33,25 @@
 	 *                        New lines can be added with "\n" or by giving multiple descriptionKeys.
 	 *                        Long lines are wrapped automatically.
 	 *                        Very long entries will span multiple pages automatically.
+	 * @deprecated Use {@link #addIngredientInfo(Object, IIngredientType, ITextComponent...)} instead.
```
mezz:
```suggestion
	 * @deprecated Since JEI 7.6.4, use {@link #addIngredientInfo(Object, IIngredientType, ITextComponent...)} instead.
```

src/api/java/mezz/jei/api/registration/IRecipeRegistration.java
```diff
@@ -45,6 +62,22 @@
 	 *                        New lines can be added with "\n" or by giving multiple descriptionKeys.
 	 *                        Long lines are wrapped automatically.
 	 *                        Very long entries will span multiple pages automatically.
+	 * @deprecated Use {@link #addIngredientInfo(List, IIngredientType, ITextComponent...)} instead.
```
mezz:
```suggestion
	 * @deprecated Since JEI 7.6.4, use {@link #addIngredientInfo(List, IIngredientType, ITextComponent...)} instead.
```

src/main/java/mezz/jei/gui/recipes/RecipesGui.java
```diff
@@ -467,9 +474,9 @@ private void updateLayout() {
 
 		logic.setRecipesPerPage(recipesPerPage);
 
-		title = recipeCategory.getTitle();
+		title = recipeCategory.getTitleAsTextComponent();
```
mezz: Does this allow font formatting like bold or colors? That's ok for tooltips, but we should strip that out of things like titles that are displayed in the JEI GUIs to enforce a more consistent look.

src/main/java/mezz/jei/plugins/jei/info/IngredientInfoRecipe.java
```diff
@@ -90,4 +89,68 @@ private IngredientInfoRecipe(List<T> ingredients, IIngredientType<T> ingredientT
...
 	}
+
+	private static class ExpandNewLineTextAcceptor implements ITextProperties.IStyledTextAcceptor<Void> {
+
+		private final List<ITextProperties> lines = new ArrayList<>();
+		@Nullable
+		private IFormattableTextComponent lastComponent;
+
+		@Override
+		public Optional<Void> accept(Style style, String line) {
```
mezz, line 100: I think this should have unit tests if possible, the rules for it are pretty complicated.

src/main/java/mezz/jei/transfer/RecipeTransferErrorTooltip.java
mezz: deepCopy is fine, I think if there is something like this that is not compiler-enforceable coming from the API then we should just be careful with it like this.

src/main/java/mezz/jei/util/StringUtil.java
```diff
@@ -1,13 +1,15 @@
...
 
 public final class StringUtil {
 	private StringUtil() {
 
 	}
 
-	public static String truncateStringToWidth(String string, int width, FontRenderer fontRenderer) {
-		return fontRenderer.getLineScrollOffset(string, width - fontRenderer.getStringWidth("..."), false) + "...";
+	public static ITextComponent truncateStringToWidth(ITextComponent text, int width, FontRenderer fontRenderer) {
+		return new StringTextComponent(fontRenderer.func_238417_a_(text, width - fontRenderer.getStringWidth("...")).getString() + "...");
```
mezz: would be nice to introduce some intermediate variables here, I try to avoid having more than one function call per line

mezz (changes requested):
Looks pretty good! I think this makes sense as a single PR and single version bump.

src/main/java/mezz/jei/plugins/jei/info/IngredientInfoRecipe.java
pupnewfster: Unfortunately the cases where the rules are complex (when lastComponent is being modified) is only really used in translation text components when doing parameter replacement and has code that ensures the game is running and the language map is initialized which makes this rather complicated to do unit tests for until 1.17 when it will hopefully become easier as from what I understand Mojang is exposing their testing suite.

src/api/java/mezz/jei/api/recipe/category/IRecipeCategory.java
mezz: The null check there is to help with broken plugins or when devs are just starting the plugin.
This is good for now, it will no longer be exposed in the API after `getTitle` is removed.

mezz (approved):

PR #2313, opened 2021-04-16
Generalize recipe name callback to work for all ingredient types

pupnewfster:
Built and tested this against Mekanism and it works for our custom ingredient types and also for fluid ingredients.


mezz (approved):
LGTM, nice!

PR #2334, opened 2021-05-06
Update FTB Library integration

LatvianModder:
It shouldn't even crash for older versions, since the mod ID changed from `ftbguilibrary` to `ftblibrary`

### DIFF
--- a/src/main/java/mezz/jei/config/JEIClientConfig.java
+++ b/src/main/java/mezz/jei/config/JEIClientConfig.java
@@ -1,7 +1,7 @@
 package mezz.jei.config;
 
-import com.feed_the_beast.mods.ftbguilibrary.config.ConfigGroup;
-import com.feed_the_beast.mods.ftbguilibrary.config.gui.GuiEditConfig;
+import dev.ftb.mods.ftblibrary.config.ConfigGroup;
+import dev.ftb.mods.ftblibrary.config.ui.EditConfigScreen;
 import mezz.jei.api.constants.ModIds;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.screen.inventory.InventoryScreen;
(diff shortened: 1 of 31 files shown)


src/main/java/mezz/jei/config/JEIClientConfig.java
```diff
@@ -70,8 +70,8 @@ public static void openSettings() {
 			};
 			gui.openGui();
 		} else {
-			mc.player.sendStatusMessage(new TranslationTextComponent(ModIds.JEI_ID + ".message.ftbguilib")
-				.setStyle(Style.EMPTY.setClickEvent(new ClickEvent(ClickEvent.Action.OPEN_URL, "https://www.curseforge.com/minecraft/mc-mods/ftb-gui-library"))), false);
+			mc.player.sendStatusMessage(new TranslationTextComponent(ModIds.JEI_ID + ".message.ftbguilib") // TODO: Change language key/value to exclude 'GUI'
```
mezz: still TODO?

mezz (changes requested):
Looks good, thanks Lat!

src/main/resources/assets/jei/lang/fr_fr.json
```diff
@@ -122,7 +122,8 @@
   "gui.jei.category.itemInformation": "Information",
 
   "_comment": "Messages",
-  "jei.message.ftbguilib": "Installez FTB GUI Library pour accéder aux configurations en jeu",
+  "jei.message.ftblibrary": "Install FTB Library to access ingame config",
+  "jei.message.ftbguilib": "Installez FTB Library pour accéder aux configurations en jeu",
```
Eric-01, line 126: Are you sure this is what it should be? ;)

mezz: oops, good catch thanks

PR #2348, opened 2021-05-12
Update fr_fr.json

Aarrn33:
Added a missing translation in french for the "jei.message.ftbguild" line.


mezz (changes requested):
Thank you for contributing!
I think the name of the mod is "FTB Library", so that part should be untranslated.

PR #2365, opened 2021-05-25
Hide empty recipe categories when viewing usages of a focus

pupnewfster:
Sparked from a discussion on the discord. Do note that I have not tested this yet as I don't personally have a major use for this.


mezz:
Thanks for contributing!

I just did a refactor here https://github.com/mezz/JustEnoughItems/commit/7413f0c6d1a3eee2582a60ad757c50af20e5e8ed
to help Jared with https://github.com/mezz/JustEnoughItems/issues/2403

I think it might also fix this issue since I unified a bunch of logic, can you confirm?

pupnewfster:
Tested by xeno on discord and it does in fact fix this.

PR #2366, opened 2021-05-25
Fix reloading not working properly on servers

### LINKED ISSUE #2302: JEI not properly updating after reload on servers
I was doing some miscellaneous testing related to using `/reload` for how my recipe lookup works/is implemented in Mekanism and noticed that JEI does not properly update on a dedicated server from `/reload` being called. I believe this was introduced by #2227 (cc: @Alwinfy ) due to the slightly incomplete implementation compared to what I described here: https://github.com/mezz/JustEnoughItems/issues/2217#issuecomment-763172521 which from my understanding is what #2227 was based on. The issue has to do with the fact that on connecting to a server for the first time the server sends the packets in this order:
- Recipes
- Tags (vanilla, then custom types)

BUT on `/reload` it sends the packets in this order:
- Tags (vanilla, then custom types)
- Recipes

The issue with the implementation of #2227 was that while it did address the fact that JEI was doing the updates in the wrong event when connecting to servers, it doesn't have any form of check to validate if the player was already connected and it should be using the other event this time. This causes it to "update" the recipes too early and be stuck with the old outdated copy.

I am not really sure of a good way to implement said fix/change (hence why I am opening it as an issue).

JEI version: 7.6.1.75
Forge version: 36.1.4

pupnewfster:
Fixes #2302

Cases tested:
- Initial join in single player
- Reloading in single player when not open to LAN
- Reloading in single player when open to LAN (both with and without another client connected)
- Initial join to a vanilla server
- Reloading on a vanilla server
- Initial join to a modded server
- Reloading on a modded server
- Joining a server leaving and joining another one

### DIFF
--- a/src/main/java/mezz/jei/startup/ClientLifecycleHandler.java
+++ b/src/main/java/mezz/jei/startup/ClientLifecycleHandler.java
@@ -30,6 +30,7 @@ import net.minecraft.client.network.play.ClientPlayNetHandler;
 import net.minecraft.network.NetworkManager;
 import net.minecraft.resources.IReloadableResourceManager;
 import net.minecraft.resources.IResourceManager;
+import net.minecraftforge.client.event.ClientPlayerNetworkEvent;
 import net.minecraftforge.client.event.RecipesUpdatedEvent;
 import net.minecraftforge.event.AddReloadListenerEvent;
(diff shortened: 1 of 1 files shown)


mezz:
Thank you for contributing! This looks good to me.

PR #2380, opened 2021-06-06
Inventory Tweaks Tree sorting 1.16

IMarvinTPA:
I hope this makes it in, but I would understand if it does not. If not, can you help me understand how to make an API that would register a new comparator? I don't think the ENum IngredientSortStage makes adding new ones easy.


Eric-01:
Isn't Inventory Tweaks discontinued (especially in 1.13+)? Wouldn't it be better to reuse sorting rules from one of the modern and still maintained mods instead, such as: [Inventory Profiles](https://github.com/jsnimda/Inventory-Profiles) (IMO the best one as it has client-side sorting), [Inventory Sorter](https://github.com/cpw/inventorysorter) (made by cpw), [Inventory Tweaks Renewed](https://github.com/reo-ar/InvTweaksRenewed), etc.?

IMarvinTPA:
I'm in the middle of adding the tree sorting feature to Inventory Tweaks Renewed, but I need to spend a fair bit of time with the network syncing code and have made no attempt yet at that.  The reason I keep adding some form of Inventory Tweaks sorting to JEI is that it is the easiest way to find a display that shows "all the things" and test out the sorter, identify what's missing and add it to the file.

mezz:
This should definitely be an API call and not all included in here. I can help with this later.

IMarvinTPA:
@mezz  Ok.  I'll make a new pull request for just the sorting config bits.  I think I'm almost done getting InventoryTweaksRenewed working properly.

Issue #2383, opened 2021-06-07
jei crash on item source  


DjBigNic977878:
[crash-2021-06-07_14.21.17-client.txt](https://github.com/mezz/JustEnoughItems/files/6612263/crash-2021-06-07_14.21.17-client.txt)
[crash-2021-06-07_15.05.41-client.txt](https://github.com/mezz/JustEnoughItems/files/6612264/crash-2021-06-07_15.05.41-client.txt)

please Help!!!

Ommina:
@IMarvinTPA Does this relate to the recent tool sorting addition?  It seems to mention it, but I can't tell if it from JEI or, as is so often the case, some other mod misbehaving and JEI getting the blame.

IMarvinTPA:
@Ommina  Yep.  I'll fix it tonight when I get back from work.

PR #2386, opened 2021-06-08
Fix crash #2383

IMarvinTPA:
Fixed the reported error [2383](https://github.com/mezz/JustEnoughItems/issues/2383?) and converted spaces to tabs to match the rest of the file.


mezz:
Gotta be super defensive with JEI, it interacts with every other mod :D

Issue #2387, opened 2021-06-08
Default sort order not working properly


Feramance:
I am playing All The Mods 6 and I tried deleting configs and allowing Minecraft to regenerate the configs however I cannot seem to get the original sort order. Would you be able to let me know how to obtain the default sort order that has been the norm for years?

Thanks,
Feramance

mezz:
Thanks for the report!

@IMarvinTPA it seems like people are having issues with the new default sort order. Can you revert the config defaults so that it is like before, at least until the in-game config is ready?

IMarvinTPA:
@Feramance Before starting the game, edit the jei-client.toml.
IngredientSortStages = ["MOD_NAME", "INGREDIENT_TYPE", "CREATIVE_MENU"]
@mezz Yeah.

Feramance:
@IMarvinTPA Thanks, I tried to mess with those myself but didn't get the correct order. Thanks for the solution. Also I agree with @mezz, better to leave the default as it was until the in-game config is ready.

Thanks

Issue #2385, opened 2021-06-08
item list in 1.16.5


nicmicir:
does the new version not list the items by item id anymore?

PR #2393, opened 2021-06-10
2387 and 2385 Original Sort Order

IMarvinTPA:
I left all of them in for visibility, but I need all of them in the list for the config screen editing which should be a very soon pull request after this one.


IMarvinTPA:
In response to #2387 and #2385.

mezz:
Thanks!

PR #2425, opened 2021-07-23
Update Mappings and FG in prep for 1.17

pupnewfster:
As requested on discord, targeting the 1.16 branch with this so that then the diff for 1.17 is smaller. A details list of changes is as follows:
- Add missing override annotations to two methods (so that when switching mappings it is easier to make sure nothing accidentally collides that didn't before).
- Switch from mcp mappings to mojmap.
- Remove two unused ATs that got left in from some previous port.
- Make the API jar get Reobf'd as there are a couple impls in it that while not needed to compile against, does make it a bit cleaner for users looking at them
- Update Forge in dev to latest RB (36.2.0)
- Update gitchangelog plugin
- Update junit
- Update to FG 5.1.+ from 4.1.+
- Update grade to 7.1.1
- Cleanup some of the warnings IntelliJ shows in the build.gradle file
- Make the replace resources task properly lazy in initialization and ensure it gets ran before classes so gradle doesn't throw warnings related to being unable to optimize builds
- Switch from `maven` to `maven-publishing` as the `maven` plugin no longer works on Gradle 7. I believe it should still work properly but there are a few small details to consider:
  1. Instead of the task being ran being called `uploadArchives` it is now `publish`
  2. It seems that each artifact is also producing an `sha256` and `sha512` file in addition to the `sha1` file
  3. The pom file does not include junit as a test thing in it (though given we don't export tests I don't think this is a problem.
  4. If desired we could add more metadata to the pom file as seen [here](https://github.com/mekanism/Mekanism/blob/cd8d57eb119b0a3bb0dc11e87a8c871cb6df5ef8/build.gradle#L625-L672) though for now I didn't bother as I wasn't sure what was desired


mezz (approved):

mezz:
Thank you!

PR #2442, opened 2021-08-03
Perform code cleanup and make use of modern Java features

pupnewfster:
See commit messages for individual changes, things of note that potentially should be considered but I didn't do for now here:
- Java now has a Set#of, List#of, and Map#ofEntries for immutable sets, lists, and maps so it may make sense in places to switch from Guava's Immutable to java's ways
- Java also now has a Stream#toList which can be used instead of .collect(Collectors.toList()) though from a brief glance I think the stream method may be immutable and create a couple extra intermediary objects unlike the old collect way so I didn't change any of those
- I didn't convert any thing to using records as those need more thought and if any of the getters override vanilla things they wouldn't work as because of the contract with how records and how forge does remapping it doesn't currently support remapping them fully yet

### DIFF
--- a/src/main/java/mezz/jei/gui/elements/DrawableAnimated.java
+++ b/src/main/java/mezz/jei/gui/elements/DrawableAnimated.java
@@ -64,20 +64,11 @@ public class DrawableAnimated implements IDrawableAnimated {
 		int animationValue = tickTimer.getValue();
 
 		switch (startDirection) {
-			case TOP:
-				maskBottom = animationValue;
-				break;
-			case BOTTOM:
-				maskTop = animationValue;
-				break;
-			case LEFT:
-				maskRight = animationValue;
-				break;
-			case RIGHT:
-				maskLeft = animationValue;
-				break;
-			default:
-				throw new IllegalStateException("Unknown startDirection " + startDirection);
+			case TOP -> maskBottom = animationValue;
+			case BOTTOM -> maskTop = animationValue;
+			case LEFT -> maskRight = animationValue;
+			case RIGHT -> maskLeft = animationValue;
+			default -> throw new IllegalStateException("Unknown startDirection " + startDirection);
 		}
 
 		drawable.draw(poseStack, xOffset, yOffset, maskTop, maskBottom, maskLeft, maskRight);
--- a/src/main/java/mezz/jei/ingredients/IngredientSorter.java
+++ b/src/main/java/mezz/jei/ingredients/IngredientSorter.java
@@ -216,7 +216,7 @@ public final class IngredientSorter implements IIngredientSorter {
 	};
 
 	private static Double getAttackDamage(ItemStack itemStack) {
-		Double attackDamage = Double.MIN_VALUE;
+		double attackDamage = Double.MIN_VALUE;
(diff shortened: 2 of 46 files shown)


src/main/java/mezz/jei/gui/elements/DrawableAnimated.java
```diff
@@ -64,20 +64,11 @@ public void draw(PoseStack poseStack, int xOffset, int yOffset) {
...
-				break;
-			case RIGHT:
-				maskLeft = animationValue;
-				break;
-			default:
-				throw new IllegalStateException("Unknown startDirection " + startDirection);
+			case TOP -> maskBottom = animationValue;
+			case BOTTOM -> maskTop = animationValue;
+			case LEFT -> maskRight = animationValue;
+			case RIGHT -> maskLeft = animationValue;
```
mezz, line 70: these lined up in a really unusual way

src/main/java/mezz/jei/ingredients/IngredientSorter.java
```diff
@@ -189,19 +189,19 @@ private static int getMaxDamage(ItemStack itemStack) {
...
-	};
+	}
 
 	private static int getHarvestLevel(ItemStack itemStack) {		
 		if (itemStack != ItemStack.EMPTY) {
 			return (itemStack.getToolTypes().size() > 0 ? itemStack.getHarvestLevel(itemStack.getToolTypes().iterator().next(), null, null) : -1);
 		}
 		return -1;
-	};
+	}
```
mezz, line 199: Looking closely, this class is pretty broken. I'll fix this logic up separately after the cleanup here.

mezz (approved):
new java is great 👍

mezz:
> Java now has a Set#of, List#of, and Map#ofEntries for immutable sets, lists, and maps so it may make sense in places to switch from Guava's Immutable to java's ways

We can explore this, but I use a lot of immutable stuff so we may run into limitations. I'd be great if everything could be native java though.


> Java also now has a Stream#toList which can be used instead of .collect(Collectors.toList()) though from a brief glance I think the stream method may be immutable and create a couple extra intermediary objects unlike the old collect way so I didn't change any of those

Could be useful, we may want to find a blog post on this to learn more since I'm sure it has some kind of purpose if they added it.

> I didn't convert any thing to using records as those need more thought and if any of the getters override vanilla things they wouldn't work as because of the contract with how records and how forge does remapping it doesn't currently support remapping them fully yet

Sounds good. There are probably some good candidates for using it but let's avoid breaking Forge!

src/main/java/mezz/jei/ingredients/IngredientSorter.java
pupnewfster: When you are doing so it may be wroth keeping in mind that the tooltype system is going away in its current form relatively soon https://github.com/MinecraftForge/MinecraftForge/pull/7970

mezz: Good to know, thanks! I just completed a quick refactor on the sorting logic, so it shouldn't be too bad to update it when that PR is merged.

PR #2449, opened 2021-08-08
Several improvements to the U.S. English translation

Eric-01:
Some changes to the U.S. English translation that should make it a little better.
if you have any comments on them, of course let me know.


mezz:
Thank you for the improvements! These look good to me.

Eric-01:
Thank you.

PR #2451, opened 2021-08-09
Replace raw Object references in the API with types where possible (includes breaking changes to API)

mezz:
This adds some more typed methods to replace ones that returned a raw `Object` ingredient.
The idea is to remove most `Object` references because they don't offer any type safety, with `Object` it's easy to make a mistake and pass in a recipe category or something else instead of a recipe or ingredient.

### DIFF
--- a/src/main/java/mezz/jei/util/ErrorUtil.java
+++ b/src/main/java/mezz/jei/util/ErrorUtil.java
@@ -269,6 +269,15 @@ public final class ErrorUtil {
 		}
 	}
 
+	public static <T> void checkIsInstance(IIngredientType<T> ingredientType, @Nullable T ingredient, String name) {
+		checkNotNull(ingredient, name);
+		Class<? extends T> ingredientClass = ingredientType.getIngredientClass();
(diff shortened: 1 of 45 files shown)


src/main/java/mezz/jei/util/ErrorUtil.java
```diff
@@ -269,6 +269,15 @@ public static void checkNotNull(@Nullable Collection<?> values, String name) {
 		}
 	}
 
+	public static <T> void checkIsInstance(IIngredientType<T> ingredientType, @Nullable T ingredient, String name) {
+		checkNotNull(ingredient, name);
```
pupnewfster: This seems like an error in that `ingredient` is annotated as nullable but if null is passed the `checkNotNull` will just cause it to error.

mezz: Good point, I've been taking the wrong approach to nullability in these `check___` methods. I'll fix them up here.

pupnewfster (approved):

PR #2469, opened 2021-08-24
Fix line rendering for ghost ingredients

shartte:


src/main/java/mezz/jei/gui/ghost/GhostIngredientDrag.java
```diff
@@ -75,22 +78,35 @@ public void drawItem(Minecraft minecraft, PoseStack poseStack, int mouseX, int m
...
-			RenderSystem.setShaderColor(red, green, blue, alpha);
-			GL11.glVertex3f(mouseX, mouseY, 150);
-			GL11.glVertex3f(originX, originY, 150);
-			GL11.glEnd();
+			builder.vertex(originX, originY, 150).color(red, green, blue, alpha).endVertex();
+			builder.vertex(mouseX, mouseY, 150).color(red, green, blue, alpha).endVertex();
+			tesselator.end();
+
+			RenderSystem.setShader(() -> oldShader);
+			RenderSystem.enableDepthTest();
```
mezz, line 95: there are 2 `enableDepthTest` in a row here, was one meant to be something else?

mezz (changes requested):
Thank you for fixing this!

src/main/java/mezz/jei/gui/ghost/GhostIngredientDrag.java
shartte: No that's just a leftover from fiddling around with it until it worked 😁 
I'll clean it up

shartte: Oh that should actually have been enableTexture! Fixed it.

src/main/java/mezz/jei/gui/ghost/GhostIngredientDrag.java
```diff
@@ -75,22 +78,35 @@ public void drawItem(Minecraft minecraft, PoseStack poseStack, int mouseX, int m
 				float percentOfDim = Math.min(1, distanceSq / (float) screenDim);
 				lineWidth = 1 + ((1 - (percentOfDim)) * 3);
 			}
-			GL11.glDisable(GL11.GL_TEXTURE_2D);
+
+			RenderSystem.disableTexture();
 			RenderSystem.disableDepthTest();
+			RenderSystem.depthMask(false);
+
+			var oldShader = RenderSystem.getShader();
```
pupnewfster, line 77: I am not sure what mezz's stance on this is, but figured I would point it out in case mezz feels strongly one way or the other and just missed it. Glancing at the JEI codebase it seems the three additions of `var` here would be the only uses of `var` in JEI.

src/main/java/mezz/jei/gui/ghost/GhostIngredientDrag.java
```diff
@@ -75,22 +78,35 @@ public void drawItem(Minecraft minecraft, PoseStack poseStack, int mouseX, int m
...
+			builder.vertex(mouseX, mouseY, 150).color(red, green, blue, alpha).endVertex();
+			tesselator.end();
+
+			RenderSystem.setShader(() -> oldShader);
 			RenderSystem.enableDepthTest();
-			GL11.glEnable(GL11.GL_TEXTURE_2D);
+			RenderSystem.enableTexture();
+			RenderSystem.depthMask(true);
+
+			GL11.glLineWidth(1);
```
pupnewfster: Is there a reason this uses a direct `GL11` call rather than `RenderSystem#lineWidth`. Same with the pre-existing lineWidth call above.

src/main/java/mezz/jei/gui/ghost/GhostIngredientDrag.java
```diff
@@ -75,22 +78,35 @@ public void drawItem(Minecraft minecraft, PoseStack poseStack, int mouseX, int m
...
+			RenderSystem.setShader(GameRenderer::getPositionColorShader);
+
 			GL11.glLineWidth(lineWidth);
 			GL11.glEnable(GL11.GL_LINE_SMOOTH);
 			GL11.glHint(GL11.GL_LINE_SMOOTH_HINT, GL11.GL_NICEST);
-			GL11.glBegin(GL11.GL_LINES);
+
+			var tesselator = RenderSystem.renderThreadTesselator();
+			var builder = tesselator.getBuilder();
+			builder.begin(VertexFormat.Mode.DEBUG_LINES, DefaultVertexFormat.POSITION_COLOR);
```
pupnewfster, line 85: Is there a meaningful difference between `Mode.DEBUG_LINES` and `Mode.LINES`. At a quick glance it looks like there might be, but I am not aware of a normal `DEBUG_LINES` GL mode so this seems to be a mojang specific thing.

src/main/java/mezz/jei/gui/ghost/GhostIngredientDrag.java
```diff
@@ -75,22 +78,35 @@ public void drawItem(Minecraft minecraft, PoseStack poseStack, int mouseX, int m
...
 			float red = (targetColor >> 24 & 255) / 255.0F;
 			float green = (targetColor >> 16 & 255) / 255.0F;
 			float blue = (targetColor >> 8 & 255) / 255.0F;
 			float alpha = (targetColor & 255) / 255.0F;
-			RenderSystem.setShaderColor(red, green, blue, alpha);
-			GL11.glVertex3f(mouseX, mouseY, 150);
-			GL11.glVertex3f(originX, originY, 150);
-			GL11.glEnd();
+			builder.vertex(originX, originY, 150).color(red, green, blue, alpha).endVertex();
+			builder.vertex(mouseX, mouseY, 150).color(red, green, blue, alpha).endVertex();
```
pupnewfster: Is there a reason this is changing it from drawing from mouse to origin, to drawing from origin to mouse? I doubt it really matters just would be probably worth mentioning in this PR why this change is being made.

pupnewfster (commented):
Oh also just tried testing this using the debug ghost handler for the brewing stand and the game's log is being spammed with this message when it is actively drawing the lines
> [Render thread/INFO]: OpenGL debug message: id=1281, source=API, type=ERROR, severity=HIGH, message='GL_INVALID_VALUE error generated. Operation is not valid from a preview context.'

src/main/java/mezz/jei/gui/ghost/GhostIngredientDrag.java
shartte: No reason, just by accident. I also don't think it matters for lines since they should not be culled in any case (i know of, at least).

src/main/java/mezz/jei/gui/ghost/GhostIngredientDrag.java
shartte: Yes. DEBUG_LINES maps to GL LINES, while Mode.LINES maps to Mojangs own line tessellation.

src/main/java/mezz/jei/gui/ghost/GhostIngredientDrag.java
shartte: I removed line width entirely now.

src/main/java/mezz/jei/gui/ghost/GhostIngredientDrag.java
shartte: Reverted this back to its original order as part of removing line width.

src/main/java/mezz/jei/gui/ghost/GhostIngredientDrag.java
mezz: Good to point out. 
I think `var` is good, I'm happy to move everything to use it.
I'm hoping Mojang's mappings remain relatively stable, for example it could be confusing for porting here if the types change and we are using `var`, since we wouldn't have much information about what type it used to be.

src/main/java/mezz/jei/gui/ghost/GhostIngredientDragManager.java
```diff
@@ -122,7 +122,28 @@ public void stopDrag() {
 	}
 
 	private <T extends Screen, V> boolean handleClickGhostIngredient(T currentScreen, IClickedIngredient<V> clicked, double mouseX, double mouseY) {
-		IGhostIngredientHandler<T> handler = guiScreenHelper.getGhostIngredientHandler(currentScreen);
+		IGhostIngredientHandler<T> handler = // guiScreenHelper.getGhostIngredientHandler(currentScreen);
```
mezz: what's all this?

mezz (changes requested):

src/main/java/mezz/jei/gui/ghost/GhostIngredientDragManager.java
shartte: Oh shit... I didn't intend to commit that 😁 
That's what I get for making that line-width change half-asleep and forgetting to exclude my "test" code.
I will remove it and re-request review....

shartte:
Alright, rebased on 1.17 branch and removed debug leftovers.

mezz (approved):

PR #2478, opened 2021-08-31
Update gradle and forge

pupnewfster:
- Updates Gradle to 7.2
- Update resource pack format to the format 1.17 uses (this was accidentally missed in the port to 1.17)
- Bumps forge version to 37.0.50 to allow for the removal of a workaround that is no longer needed in the `DebugRecipeCategory`. This also includes requiring at least forge version 37.0.47 which adjusts how custom tag types are handled, which JEI already accommodated for in the port to 1.17. The decision to accommodate for it ahead of time was so as to not to have JEI break when the PR was merged given custom tags didn't really work in 1.17 before forge 37.0.47, and the main consumers of them that use them in recipes that would show up in JEI are staying on 1.16 for a while longer. Requiring a version that contains the changes of 37.0.47 will make sure that JEI does load properly even if any mods try and use custom tag types.


mezz:
Looks good, thank you!

PR #2530, opened 2021-10-19
Update zh_tw.json

HJ-zhtw:
Update notes:
* fa0e1e8f1b25cfdb21e679c0497f0c604205499c
* Used "封鎖" (Block) instead of "黑" (Black) on line 93 and 94.


mezz:
Thank you!

PR #2587, opened 2021-12-07
Update de_de.json

Wizqok:
matches en_us now


mezz:
Thank you for contributing!

PR #2629, opened 2021-12-29
Custom Sorting Per-Item List

Nekkok:
Add the option to create custom per-item list order. I thinks is really necessary if you want to create a logical order list with different items from different mods and would be really helpful to create modpacks.


mezz:
Sorry for this type of request please open a new issue, not a pull request.

PR #2666, opened 2022-01-08
Use NBTIngredient for the potion in generated tipped arrow recipes

shartte:
Use NBTIngredient for the potion in generated tipped arrow recipes to allow recipe transfer handlers to use normal match/ingredient.test logic.


mezz:
Thanks for the fix!

mont-40:
Causes #2696

PR #2701, opened 2022-01-29
Refactor: make it easier to set recipe ingredients in the API

mezz:


src/main/java/mezz/jei/util/TagUtil.java
```diff
@@ -0,0 +1,35 @@
...
+import net.minecraft.world.item.ItemStack;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+
+public class TagUtil {
+	@Nullable
+	public static ResourceLocation getTagEquivalent(Collection<@Nullable ItemStack> itemStacks) {
```
pupnewfster: I wonder if this method could be improved at all. I started trying to make some improvements on a branch of mine in this commit https://github.com/pupnewfster/JustEnoughItems/commit/d30f9c4f98da5db941389087a3ec48c1de378a89, which makes it support showing tag equivalents for fluids, but I wasn't sure how to best do it from an API perspective for custom ingredients (such as Mekanism's chemicals) AND also the fact that a lot of recipes that add fluids filter out displaying any "flowing" fluids that are usually in the tag as well.

src/api/java/mezz/jei/api/ingredients/IIngredientRenderer.java
```diff
@@ -24,9 +25,16 @@
 	 *
 	 * @param xPosition  The x position to render the ingredient.
 	 * @param yPosition  The y position to render the ingredient.
-	 * @param ingredient the ingredient to render. May be null, some renderers (like fluid tanks) will render a
+	 * @param width      The width to render the ingredient with.
+	 * @param height  	 The height to render the ingredient with.
```
pupnewfster: Spacing seems somewhat messed up here

src/api/java/mezz/jei/api/ingredients/ITypedIngredient.java
```diff
@@ -0,0 +1,9 @@
+package mezz.jei.api.ingredients;
+
+/**
+ * Ingredient with type information, for type safety.
+ */
+public interface ITypedIngredient<T> {
```
pupnewfster, line 9: This interface with how straightforward it is almost feels like it would make more sense as a record, or at least with having the method names as `type()` and `ingredient()` to allow for making the internal implementation a record instead of a final class.

src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferHandler.java
```diff
@@ -40,7 +57,11 @@
 	 * @return a recipe transfer error if the recipe can't be transferred. Return null on success.
 	 *
 	 * @since JEI 7.1.3
+	 *
+	 * @deprecated since JEI 9.3.0.
+	 * Use {@link #transferRecipe(AbstractContainerMenu, Object, IRecipeSlotsView, Player, boolean, boolean)} instead.
 	 */
+	@Deprecated(forRemoval = true, since = "9.3.0")
 	@Nullable
 	default IRecipeTransferError transferRecipe(C container, R recipe, IRecipeLayout recipeLayout, Player player, boolean maxTransfer, boolean doTransfer) {
```
pupnewfster, line 65: Just putting a note so it doesn't get lost but this deprecated version doesn't seem to be called anymore so any implementations would require an update.

src/api/java/mezz/jei/api/recipe/category/IRecipeCategory.java
```diff
@@ -50,53 +53,60 @@
...
+	 * @param mouseX          the X position of the mouse, relative to the recipe.
+	 * @param mouseY          the Y position of the mouse, relative to the recipe.
 	 * @return tooltip strings. If there is no tooltip at this position, return an empty list.
+	 *
+	 * @since JEI 9.3.0
 	 */
-	default List<Component> getTooltipStrings(T recipe, double mouseX, double mouseY) {
-		return Collections.emptyList();
+	default List<Component> getTooltipStrings(T recipe, IRecipeSlotsView recipeSlotsView, double mouseX, double mouseY) {
+		// if not implemented, this calls the old draw function for backward compatibility
```
pupnewfster: Looks like a copy paste error, should be calls the old getTooltipStrings function

src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferHandler.java
mezz: good catch, I have changed this to fix backward compatibility

src/api/java/mezz/jei/api/ingredients/ITypedIngredient.java
mezz: I like keeping the interface for the API, this structure has served me very well for refactors!
I think making this a record makes sense, maybe I'll just add some legacy getters to it to keep the `get` convention consistent.

mezz: Hm actually the `TypedIngredient` implementation is pretty complex, with a private constructor. I'm not sure records work in this case.

src/main/java/mezz/jei/util/TagUtil.java
mezz: To make it general, I think this could be offloaded completely into something like 
`IIngredientHelper#getTagEquivalent(Collection<V> ingredients)`

The fluid handling is complicated so I've stuck with items only so far. There may be a good way to handle fluids in a custom way in its own `IIngredientHelper`.

src/api/java/mezz/jei/api/gui/ingredient/IRecipeSlotsView.java
```diff
@@ -0,0 +1,43 @@
...
+ * Useful for implementing {@link IRecipeTransferHandler} and some other advanced cases.
+ *
+ * This view is meant to replace {@link IRecipeLayout} and {@link IGuiIngredientGroup}
+ * as a source of information for recipe transfer, drawing, and tooltips.
+ *
+ * @since JEI 9.3.0
+ */
+public interface IRecipeSlotsView {
+	/**
+	 * Get the list of slots that have ingredients in them for a recipe.
```
pupnewfster: This and the other `getSlotViews` method potentially should have some other wording given it is no longer a "list". Maybe just "Get the slots ..."

src/api/java/mezz/jei/api/gui/builder/IRecipeSlotBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+
+	/**
+	 * Set the rendering size of the ingredient and its background.
+	 *
+	 * @param width  the full width of the rendered ingredient and its background
+	 * @param height the full height of the rendered ingredient and its background
+	 *
+	 * @since JEI 9.3.0
+	 */
+	IRecipeSlotBuilder setSize(int width, int height);
```
pupnewfster: I feel like the javadocs for this method should clarify how width/height is handled when there is padding. Is it the width/height of the total thing and then the renderer gets called with it automatically being adjusted by the inner padding or does it not actually include the background in the width/height if there is padding taking place.

src/api/java/mezz/jei/api/gui/builder/IRecipeLayoutBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+	 */
+	void moveRecipeTransferButton(int posX, int posY);
+
+	/**
+	 * Adds a shapeless icon to the top right of the recipe,
+	 * that shows a tooltip saying "shapeless" when hovered over.
+	 *
+	 * @since JEI 9.3.0
+	 */
+	void setShapeless();
```
pupnewfster, line 60: This is nice and for me fits perfectly into the GUI, but I wonder if it would make sense to be able to also add a moveShapelessIcon in case it is in a bad spot. Also unlike the javadoc says it doesn't seem like this actually shows a tooltip saying shapeless when hovering over the icon.

src/api/java/mezz/jei/api/ingredients/ITypedIngredient.java
pupnewfster: Yeah that makes sense, I was just bringing it up initially because interface wise the class seemed pretty straightforward.

src/api/java/mezz/jei/api/gui/builder/IRecipeLayoutBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+	 * Add a slot that will be drawn at the given position relative to the recipe layout.
+	 *
+	 * @param recipeIngredientRole the {@link RecipeIngredientRole} of this slot (for lookups).
+	 * @param x relative x position of the slot on the recipe layout.
+	 * @param y relative y position of the slot on the recipe layout.
+	 * @return a {@link IRecipeSlotBuilder} that has further methods for adding ingredients, etc.
+	 *
+	 * @since JEI 9.3.0
+	 */
+	IRecipeSlotBuilder addSlot(RecipeIngredientRole recipeIngredientRole, int x, int y);
```
pupnewfster, line 34: This wasn't a big issue for me but I wonder if it is worth pointing out that this x and y is off by one compared to the old system when dealing with items (as the old one automatically padded by 1, 1 for slots) so if someone isn't testing their stuff or putting much thought into it they may run into by accident. (JEI's own categories actually suffer from this issue)

Personally I think it is better to not have the automatic magic padding logic included, but I do think it is worth a mention in the javadocs.

src/api/java/mezz/jei/api/recipe/transfer/IRecipeTransferHandlerHelper.java
```diff
@@ -31,8 +33,26 @@
...
 	 * @param tooltipMessage   the message to show on the tooltip for the recipe transfer button.
 	 * @param missingItemSlots the slot indexes for items that are missing. Must not be empty.
-	 *                         Slots are indexed according to {@link IGuiItemStackGroup#getGuiIngredients()}.
+	 *
+	 * @since JEI 9.3.0
+	 */
+	IRecipeTransferError createUserErrorForMissingSlots(
+		Component tooltipMessage,
+		Collection<IRecipeSlotView> missingItemSlots
+	);
```
pupnewfster: This feels a little odd being across multiple lines given it doesn't seem that much longer than the method being deprecated.

src/main/java/mezz/jei/collect/MultiMap.java
```diff
@@ -22,12 +24,17 @@ public MultiMap(Map<K, T> map, Supplier<T> collectionSupplier) {
...
 	}
 
-	public T get(K key) {
-		return map.computeIfAbsent(key, collectionMappingFunction);
+	public Collection<V> get(K key) {
+		T collection = map.get(key);
+		if (collection != null) {
+			return Collections.unmodifiableCollection(collection);
+		}
+		return List.of();
```
pupnewfster: Doesn't really matter but this might be more logical as `Collections#emptyList()` rather than importing `List` just to use this when `Collections` is already being imported for `unmodifiableCollection`

src/main/java/mezz/jei/plugins/vanilla/ingredients/fluid/FluidStackRenderer.java
```diff
@@ -82,10 +88,16 @@ public void render(PoseStack poseStack, final int xPosition, final int yPosition
...
 
-	private void drawFluid(PoseStack poseStack, final int xPosition, final int yPosition, @Nullable FluidStack fluidStack) {
-		if (fluidStack == null) {
+	@SuppressWarnings("removal")
+	@Override
+	public void render(PoseStack stack, int xPosition, int yPosition, @Nullable FluidStack ingredient) {
+		if (ingredient == null) {
 			return;
 		}
+		render(stack, xPosition, yPosition, 16, 16, ingredient);
```
pupnewfster: I feel like at this point it would be more readable to do:
```java
if (ingredient != null) {
    render(stack, xPosition, yPosition, 16, 16, ingredient);
}
```
rather than having the early return when the block of code afterwards is one line long.

src/main/java/mezz/jei/recipes/IngredientToRecipesMap.java
```diff
@@ -0,0 +1,29 @@
...
+			List<R> recipes = uidToRecipes.computeIfAbsent(uid, k -> new ArrayList<>());
+			recipes.add(recipe);
+		}
+	}
+
+	@UnmodifiableView
+	public List<R> get(String ingredientUid) {
+		List<R> recipes = uidToRecipes.get(ingredientUid);
+		if (recipes == null) {
+			return List.of();
```
pupnewfster: Same comment here about `List#of` vs `Collections#emptyList`

src/main/java/mezz/jei/plugins/vanilla/brewing/BrewingRecipeUtil.java
```diff
@@ -52,29 +52,27 @@ private void clearCache() {
...
 		}
+
+		Collection<String> prevPotions = potionMap.get(potionOutputUid);
+		int minPrevSteps = prevPotions.stream()
+			.mapToInt(prevPotion -> getBrewingSteps(prevPotion, previousSteps))
+			.min()
+			.orElse(Integer.MAX_VALUE);
+
+		if (minPrevSteps == Integer.MAX_VALUE) {
+			return Integer.MAX_VALUE;
```
pupnewfster: Should this cache max int as the number of steps required so that it doesn't have to recalculate that value each time?

src/main/java/mezz/jei/transfer/BasicRecipeTransferHandlerServer.java
```diff
@@ -1,281 +1,358 @@
...
-
-		// grab items from slots
-		Map<Integer, ItemStack> slotMap = new HashMap<>(slotIdMap.size());
-		for (Map.Entry<Integer, Integer> entry : slotIdMap.entrySet()) {
-			Slot slot = container.getSlot(entry.getValue());
-			final ItemStack slotStack = slot.getItem();
-			if (slotStack.isEmpty()) {
+	public static void setItems(
+		Player player,
+		List<Tuple<Slot, Slot>> recipeSlotToSourceSlots,
```
pupnewfster: I feel like this method would be more readable in intentions by making this a list of a record instead of just using a Tuple

src/main/java/mezz/jei/transfer/BasicRecipeTransferHandlerServer.java
```diff
@@ -1,281 +1,358 @@
...
+	@Nullable
+	private static Map<Slot, ItemStack> calculateRequiredStacks(List<Tuple<Slot, Slot>> recipeSlotToSourceSlots, Player player) {
+		Map<Slot, ItemStack> recipeSlotToRequired = new HashMap<>(recipeSlotToSourceSlots.size());
+		for (Tuple<Slot, Slot> entry : recipeSlotToSourceSlots) {
+			Slot recipeSlot = entry.getA();
+			Slot inventorySlot = entry.getB();
+			if (!inventorySlot.mayPickup(player)) {
+				LOGGER.error(
+					"Tried to transfer recipe but was given an" +
+						" inventory slot that the player can't pickup from: " + inventorySlot.index);
```
pupnewfster: Shouldn't this be using the `{}` replace syntax for the slot index?

src/main/java/mezz/jei/transfer/BasicRecipeTransferHandlerServer.java
```diff
@@ -1,281 +1,358 @@
...
+				return null;
+			}
+			final ItemStack slotStack = inventorySlot.getItem();
+			if (slotStack.isEmpty()) {
+				LOGGER.error(
+					"Tried to transfer recipe but was given an" +
+						" empty inventory slot as an ingredient source: " + inventorySlot.index);
```
pupnewfster: Shouldn't this be using the `{}` replace syntax for the slot index?

src/main/java/mezz/jei/transfer/BasicRecipeTransferHandlerServer.java
```diff
@@ -1,281 +1,358 @@
...
+				return null;
 			}
+			ItemStack stack = slotStack.copy();
+			stack.setCount(1);
```
pupnewfster, line 158:
```java
ItemStack stack = ItemHandlerHelper.copyStackWithSize(slotStack, 1);
```

src/main/java/mezz/jei/gui/recipes/OutputSlotTooltipCallback.java
```diff
@@ -0,0 +1,69 @@
...
+
+	@Override
+	public void onTooltip(IRecipeSlotView recipeSlotView, List<Component> tooltip) {
+		Optional<ITypedIngredient<?>> displayedIngredient = recipeSlotView.getDisplayedIngredient();
+		if (displayedIngredient.isEmpty()) {
+			return;
+		}
+		if (recipeSlotView.getRole() != RecipeIngredientRole.OUTPUT) {
+			return;
+		}
```
pupnewfster: I feel like this check potentially should be done before checking if the displayed ingredient is empty as it is a more straightforward check and if it isn't an output there is no reason to bother checking which ingredient is currently displayed if any.

src/main/java/mezz/jei/ingredients/IngredientAcceptor.java
```diff
@@ -0,0 +1,86 @@
...
+
+	@Override
+	public IngredientAcceptor addIngredientsUntyped(List<?> ingredients) {
+		Preconditions.checkNotNull(ingredients, "ingredients");
+
+		for (Object ingredient : ingredients) {
+			Optional<ITypedIngredient<?>> typedIngredient = TypedIngredient.create(this.ingredientManager, ingredient);
+			typedIngredient.ifPresent(i -> this.types.add(i.getType()));
+
+			this.ingredients.add(typedIngredient);
```
pupnewfster, line 41: It feels like adding to this if `typeIngredient` isn't present and is an empty optional is an oversight.

src/main/java/mezz/jei/ingredients/IngredientAcceptor.java
```diff
@@ -0,0 +1,86 @@
...
+		ErrorUtil.checkNotNull(ingredient, "ingredient");
+
+		addIngredientInternal(ingredientType, ingredient);
+		return this;
+	}
+
+	private <T> void addIngredientInternal(IIngredientType<T> ingredientType, @Nullable T ingredient) {
+		Optional<ITypedIngredient<?>> typedIngredient = TypedIngredient.create(this.ingredientManager, ingredientType, ingredient);
+		typedIngredient.ifPresent(i -> this.types.add(i.getType()));
+		this.ingredients.add(typedIngredient);
```
pupnewfster, line 71: It feels like adding to this if typeIngredient isn't present and is an empty optional is an oversight.

src/main/java/mezz/jei/ingredients/IngredientManager.java
```diff
@@ -45,10 +49,9 @@ public IngredientManager(
 		}
 		this.ingredientsMap = ingredientsMapBuilder.build();
 
-		//noinspection UnstableApiUsage
 		this.registeredIngredientTypes = ingredientInfos.stream()
 			.map(IngredientInfo::getIngredientType)
-			.collect(ImmutableList.toImmutableList());
+			.collect(Collectors.toList());
```
pupnewfster: Shouldn't this be `.toList()` instead of `.collect(Collectors.toList())` to maintain immutability of the list?

src/api/java/mezz/jei/api/gui/builder/IRecipeSlotBuilder.java
mezz: Thanks for flagging this!
This functionality is actually very confusing and I have to fix the padding of all my plugins... will put fixing this comment on the TODO before merging.

src/api/java/mezz/jei/api/gui/builder/IRecipeLayoutBuilder.java
mezz: Sounds like a good simple improvement.

src/api/java/mezz/jei/api/gui/builder/IRecipeLayoutBuilder.java
mezz: I definitely want to move away from the magic 1px padding. 
I will try to come up with something that makes it easier to understand and then I can add it to the javadocs.

src/main/java/mezz/jei/transfer/BasicRecipeTransferHandlerServer.java
mezz: Good idea, I actually changed this to a list of `TransferOperation` records before I read your comment heh.

src/main/java/mezz/jei/transfer/BasicRecipeTransferHandlerServer.java
mezz: It would look cleaner, but I don't see much use in relying on Forge in this specific case. 
Forge methods are moved or removed pretty frequently, and this one is not overridable or hooked into so it's not adding much benefit.

src/main/java/mezz/jei/ingredients/IngredientAcceptor.java
mezz: `this.ingredients` holds `Optional<ITypedIngredient<?>>` because lists of displayed ingredients are allowed to contain "blank" ingredients in the rotation (displaying nothing). I represent that here with `Optional.empty()` when it used to be a very sneaky `null`.
I'll add some clarifying doc comments to `this.ingredients`

src/main/java/mezz/jei/ingredients/IngredientAcceptor.java
mezz: see comment on https://github.com/mezz/JustEnoughItems/pull/2701#discussion_r800996702

src/main/java/mezz/jei/ingredients/IngredientManager.java
mezz: I have some weird cases where the stream types become complicated because of generics.

One example:
Given these types:
```java
// List<ITypedIngredient<?>> matches;
// List<Optional<ITypedIngredient<?>>> displayIngredients;
```

this way
```java
displayIngredients = matches.stream()
				.map(Optional::of)
				.collect(Collectors.toList());
```
works,

but if I use `.toList()` it needs some explicit casting:
```java
displayIngredients = matches.stream()
				.<Optional<ITypedIngredient<?>>>map(Optional::of)
				.toList();
```

I'll see if I can figure out the casting on more of these because it seems beneficial to make them properly immutable.

src/main/java/mezz/jei/plugins/vanilla/brewing/BrewingRecipeUtil.java
mezz: thanks, makes sense

src/main/java/mezz/jei/transfer/BasicRecipeTransferHandlerServer.java
pupnewfster: That is fair enough, mostly wanted to point out that method exists.

src/main/java/mezz/jei/ingredients/IngredientAcceptor.java
pupnewfster: Ah okay, that makes more sense.

src/main/java/mezz/jei/transfer/TransferOperation.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Represents transferring an ItemStack from inventorySlot to craftingSlot.
+ */
+public record TransferOperation(Slot inventorySlot, Slot craftingSlot) {
+	public static TransferOperation readPacketData(FriendlyByteBuf buf, AbstractContainerMenu container) {
+		int inventorySlotIndex = buf.readVarInt();
+		int craftingSlotIndex = buf.readVarInt();
+
+		Slot inventorySlot = container.getSlot(inventorySlotIndex);
+		Slot craftingSlot = container.getSlot(craftingSlotIndex);
```
pupnewfster, line 16: Especially if the client sends back indices getSlot may return null, so might be worth sanitizing that the slots exist and it isn't just a modified JEI instance sending bad data to try and crash the server.

src/main/java/mezz/jei/transfer/RecipeTransferUtil.java
```diff
@@ -67,66 +71,158 @@ public static boolean transferRecipe(RecipeTransferManager recipeTransferManager
...
+			}
+		}
+
+		// check that all slots can be picked up by the player
+		{
+			List<Integer> invalidPickupSlots = Stream.concat(
+					craftingSlots.stream(),
+					inventorySlots.stream()
+				)
+				.filter(slot -> !slot.mayPickup(player))
```
pupnewfster, line 161: This check is technically a breaking change as it causes one of the containers I have an `IRecipeTransferInfo` to no longer work as if the slot is empty I return false for `mayPickup`. Previously (and in the remaining places in `BasicRecipeTransferHandler`, though I am slightly unsure about the Server calls) before checking and failing if unable to pickup from a slot it was checked that the slot is not empty.

src/main/java/mezz/jei/transfer/BasicRecipeTransferHandlerServer.java
mezz: Thanks!

src/main/java/mezz/jei/transfer/RecipeTransferUtil.java
mezz: That makes sense. I will change this so that it only checks `mayPickup` on slots that contain something.

src/main/java/mezz/jei/transfer/TransferOperation.java
mezz: JEI's server packet handlers are pretty safe, wrapped in a try/catch/log here:
https://github.com/mezz/JustEnoughItems/blob/2df668b5ac4a8473b9837ad2785d0f5a4fb845a6/src/main/java/mezz/jei/network/PacketHandler.java#L43-L50

pupnewfster: Ah okay

src/api/java/mezz/jei/api/gui/builder/IRecipeLayoutBuilder.java
mezz: added the new method, and fixed the tooltip

src/api/java/mezz/jei/api/gui/builder/IRecipeLayoutBuilder.java
mezz: Pushed some changes for this.
Porting will require adding +1 to a bunch of coordinates, but it should be worth it to have things unified and less magical.
I simplified the way padding works and made padding into an offset that is specific to the background and overlay.
Now the position of the ingredient is the reference point, and you can offset the background by -1 for example.

src/api/java/mezz/jei/api/gui/builder/IRecipeSlotBuilder.java
mezz: I have changed this size to be per-IngredientRenderer, which I believe turns out to make it a lot simpler.

src/main/java/mezz/jei/util/TagUtil.java
mezz: Added this as `IIngredientHelper#getTagEquivalent`

src/main/java/mezz/jei/gui/ingredients/RecipeSlot.java
```diff
@@ -0,0 +1,312 @@
...
+				background.draw(poseStack);
+			}
+
+			RenderSystem.enableBlend();
+
+			getDisplayedIngredient()
+				.ifPresent(ingredient -> drawIngredient(poseStack, ingredient));
+
+			if (overlay != null) {
+				overlay.draw(poseStack);
```
pupnewfster: Similarly to how the overlay used to work in `FluidStackRenderer`, this needs to be shifted forwards by 200 in the z direction to ensure it renders over the displayed ingredient. Also I am unsure if it is intended or not, but given at the end of `FluidStackRenderer` blend is disabled that means this overlay is being drawn with blend disabled instead of enabled similarly to how it used to be drawn when drawn as part of `FluidStackRenderer`

src/main/java/mezz/jei/gui/recipes/builder/RecipeSlotBuilder.java
```diff
@@ -0,0 +1,168 @@
...
+	public IRecipeSlotBuilder setBackground(IDrawable background, int xOffset, int yOffset) {
+		ErrorUtil.checkNotNull(background, "background");
+		this.background = new OffsetDrawable(background, xOffset, yOffset);
+		return this;
+	}
+
+	@Override
+	public IRecipeSlotBuilder setOverlay(IDrawable overlay, int xOffset, int yOffset) {
+		ErrorUtil.checkNotNull(overlay, "overlay");
+		this.overlay = new OffsetDrawable(overlay, xOffset, yOffset);
```
pupnewfster: Part of me wonders (for both this and `setBackground`) if `xOffset` and `yOffset` are both zero should it just use the passed in overlay instead of wrapping it in an `OffsetDrawable`

src/main/java/mezz/jei/gui/ingredients/RecipeSlot.java
mezz: This worked for me by chance, but I have added those things back to make sure it works consistently.

src/main/java/mezz/jei/gui/recipes/builder/RecipeSlotBuilder.java
mezz: Sounds good, I'll change this to some kind of `OffsetDrawable.create` that can pass it through if there's no offset.

PR #2734, opened 2022-02-27
Improve performance and memory size of the ingredient filter

mezz:
`GeneralizedSuffixTree ` generates and stores a ton of `String#substring` fragments, which bloats memory.
This PR creates the `SubString` which points to a string and keeps an `offset` and `length` in order to act like a substring view of a string. This means we can use 1 string and point to it instead of copying it tons of times and making tiny pieces.

I also did a refactor of `GeneralizedSuffixTree` and tried to clarify the code. It's very hard to understand, hopefully this helps.
Added a bunch of tests as well, every time I touched anything I would break it horribly until I added the tests.

I also created a `LimitedStringStorage` wrapper for `GeneralizedSuffixTree`, which optimizes cases like "mod name search" where you have just a handful of strings but many many values. It groups the values into sets and manipulates those direectly, so that there is no giant duplication of values in the `GeneralizedSuffixTree`.


mezz:
Testing on ATM7 (big mod pack). 

### Load time 
Load time is slightly faster.

* Latest released version of JEI: 
```
[15:46:39] [Render thread/INFO]: Building ingredient filter...
[15:46:40] [Render thread/INFO]: Building ingredient filter took 1.132 s
```

* At current commit:
```
[20:45:03] [Render thread/INFO]: Building ingredient filter...
[20:45:03] [Render thread/INFO]: Adding 19023 ingredients
[20:45:04] [Render thread/INFO]: Added 19023 ingredients
[20:45:04] [Render thread/INFO]: Building ingredient filter took 900.5 ms
```


### Memory
Memory savings is dramatic:

* Latest released version of JEI: 
`ElementSearch 113898888 bytes` (108.52 mebibytes)

* At current commit:
`ElementSearch 42819488 bytes` (40.84 mebibytes)

PR #2739, opened 2022-03-02
Port to 1.18.2

pupnewfster:
Initial port to 1.18.2 that seems to work fine. I didn't test the ingredient sort comparator as I was not quite sure how to best do so. I also changed the suspicious stew recipe creator to more closely mimic the recipe itself as it is based on the item tag not the block tag. For vanilla flowers this shouldn't make a difference but for purposes of modded flowers where maybe the flower is only added to one of the tags this fixes the discrepancy.

I am also slightly unsure if accessing things via the vanilla registry is the correct way to do things but it seems like that is how things are done now and there doesn't appear to be a way to do so (at least yet) via the forge registry.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,18 +1,18 @@
 org.gradle.daemon=false
 
 # Minecraft
-minecraft_version=1.18.1
-minecraft_version_range=[1.18.1, 1.19)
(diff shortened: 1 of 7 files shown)


mezz:
Thanks!

PR #2757, opened 2022-03-15
Fix/kotlin build gradle

jaredlll08:


jaredlll08:
Closed due to targeting the wrong branch

PR #2760, opened 2022-03-16
use kotlin for build.gradle

mezz:


mezz:
replaced by https://github.com/mezz/JustEnoughItems/pull/2761

mezz:
Leaving this untouched so that it can be looked at for https://youtrack.jetbrains.com/issue/KTIJ-21353

PR #2777, opened 2022-03-26
Update README.md

nayfaan:
I think it would be good to clearly indicate on the front page that the mod is built for forge (and not fabric) to prevent confusion.


nayfaan:
and also on the curseforge page too, if possible.

mezz:
Thanks for contributing! 
JEI is available for Fabric now so I will close this.

PR #2842, opened 2022-06-08
JEI 10. Add Fabric support. Refactor into a multi-project

mezz:


Common/src/main/java/mezz/jei/common/color/ColorName.java
```diff
@@ -0,0 +1,4 @@
+package mezz.jei.common.color;
+
+public record ColorName(String name, Integer color) {
```
pupnewfster: Is there a reason for having the color be an `Integer` as opposed to an `int`?

mezz: hm no good reason, I should change it

PR #2854, opened 2022-06-16
[1.19] Fix improper property use causing invalid dependency POM

sciwhiz12:
The `archivesName` property of `BasePluginExtension` is a `Property<String>`. The default implementation of `toString()` for a `Property` is to print out their name and owning object, and not the `toString()` of their contents.

This meant that passing `archivesName` to `Node#appendNode` as a value, which calls `toString()` on it to make it into a String, results in e.g. `extension 'base' property 'archivesName'` instead of the expected value of e.g. `jei-1.19-common-api`.

The simple fix is to call `Property#get()` on `archivesName` to get its actual value and pass that on as the node value.

### DIFF
--- a/FabricApi/build.gradle.kts
+++ b/FabricApi/build.gradle.kts
@@ -103,7 +103,7 @@ publishing {
                 dependencyProjects.forEach {
                     val dependencyNode = dependenciesNode.appendNode("dependency")
                     dependencyNode.appendNode("groupId", it.group)
-                    dependencyNode.appendNode("artifactId", it.base.archivesName)
+                    dependencyNode.appendNode("artifactId", it.base.archivesName.get())
                     dependencyNode.appendNode("version", it.version)
                 }
             }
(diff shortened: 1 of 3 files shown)


mezz:
Thank you!

mezz:
Oops, looks like CLA bot was very late to reply to this one. Can you please sign the CLA?

sciwhiz12:
It seems that I am somehow unable to sign the CLA:
image

mezz:
I guess they are having issues right now, that's concerning. 
I can ping you later to try again.

mezz:
@sciwhiz12 I left the page open for 10+ minutes and it looks like it finally loaded, maybe that'll work for you?

It looks like they're focusing on developing a new Github Actions system for CLAs that may work better, I'll probably want to switch over to that when it's ready.

sciwhiz12:
Signed!

mezz:
Thanks!

PR #2886, opened 2022-07-12
Update to support the breaking changes in forge

pupnewfster:
This PR bumps the min forge version to 41.0.94.  I am not quite happy with how I got key bindings working and being registered but I wasn't really sure of a better way to handle it (I am also not 100% positive I didn't somehow break fabric but in theory it should work and compile still)

### DIFF
--- a/Common/src/main/java/mezz/jei/common/config/KeyBindings.java
+++ b/Common/src/main/java/mezz/jei/common/config/KeyBindings.java
@@ -1,5 +1,6 @@
 package mezz.jei.common.config;
 
+import java.util.function.Consumer;
 import mezz.jei.common.input.IKeyBindings;
 import mezz.jei.common.input.keys.IJeiKeyMapping;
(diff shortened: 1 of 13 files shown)


sekwah41:
Given the changes you've done the most likely place, you would be to break fabric I think would be the keybinds. I know when I was tinkering with forge its quite easy to have code that compiles but just doesn't register the keybinds as it's not called right.

Otherwise, if it compiles it looks like most of the rest should be fine.

If you can just give it a quick test to see if the keybinds register properly on fabric.

pupnewfster:
I have never been able to get jei to compile on fabric even with fresh clones. I always end up having to comment it out in the settings gradle in order to build and test forge. But given I am just adding consumer's to method signatures I don't see how it could break.

mezz:
This looks good to me, it doesn't really change how the Fabric keybinds work at all.

mezz:
Thank you!

PR #2971, opened 2022-08-30
Allow recipe transfer errors to set button color highlight

Technici4n:
I modified the `RecipeTransferErrorTooltip` to test this.

- **Use case 1**: render useful info but keep the normal display of the button (by returning `0`). Can also be used to change the tooltip message (for example `Encode pattern` instead of `Move items` for AE2).
image
- **Use case 2**: change color of the button to indicate different behavior. For example, in AE2, I will use orange when some items are missing, transparent when everything is available, and blue when everything is available or craftable. Image when I use `return 0x804545FF;`:
image

### DIFF
--- a/CommonApi/src/main/java/mezz/jei/api/recipe/transfer/IRecipeTransferError.java
+++ b/CommonApi/src/main/java/mezz/jei/api/recipe/transfer/IRecipeTransferError.java
@@ -37,7 +37,7 @@ public interface IRecipeTransferError {
 	Type getType();
 
 	/**
-	 * Return the RGBA color of the additional button highlight for {@link Type#COSMETIC}.
+	 * Return the ARGB color of the additional button highlight for {@link Type#COSMETIC}.
 	 * For example, return 0 to disable the colored highlight. Default color is orange.
 	 *
 	 * @since 11.2.1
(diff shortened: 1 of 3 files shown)


CommonApi/src/main/java/mezz/jei/api/recipe/transfer/IRecipeTransferError.java
```diff
@@ -36,6 +36,14 @@ enum Type {
 
 	Type getType();
 
+	/**
+	 * Return the RGBA color of the additional button highlight for {@link Type#COSMETIC}.
+	 * For example, return 0 to disable the colored highlight. Default color is orange.
+	 */
```
mezz, line 44: Please bump the version in `gradle.properties`
and add to the javadoc here
`@since <version>`
So that people know to set a minimum required version if they use this method.

mezz (changes requested):
looks good, just needs some bookkeeping to make sure users end up with the right version

mezz:
Just a note on your implementation in AE:
I think it would be helpful to show a tooltip to explain the different colors to users. Maybe you do that already but it's not visible in your screenshot so I thought I'd mention it.

CommonApi/src/main/java/mezz/jei/api/recipe/transfer/IRecipeTransferError.java
```diff
@@ -36,6 +36,16 @@ enum Type {
 
 	Type getType();
 
+	/**
+	 * Return the RGBA color of the additional button highlight for {@link Type#COSMETIC}.
```
mezz: I think this is not RGBA, or else A would be 00 and it wouldn't be visible.
ARGB probably?

mezz (changes requested):

mezz (approved):

mezz:
Thanks for contributing!

PR #2972, opened 2022-08-31
Make the `Show Recipes` tooltip optional in the API

### LINKED ISSUE #2794: [Suggestion]: expose API endpoint to disable default `Show Recipes` tooltip
### Suggestion

I'd like to handle the tooltip when hovering over the click area myself because I display a lot of information on the same spot.
Since it's very conventional that the recipe click area is the progress arrow, I would have two overlapping tooltips.

Registering a custom GUI handler through `IGuiHandlerRegistration#addGuiContainerHandler` is possible and I am able to overwrite `IGuiClickableArea#getTooltipStrings` but this method won't allow you to disable the tooltip rendering at all. If you return an empty list, it will use the default `Show Recipes` tooltip which is quite annoying.

I suggest that the tooltip is disabled completely if you return an empty list and there is another default implementation that (if not overridden) still shows the default tooltip.
Or maybe expose another method like `tooltipEnabled` which returns a boolean.

rlnt:
fixes #2794

This PR adds the ability to disable the default `Show Recipes` tooltip on the `IGuiClickableArea`.

Since the default implementation returns `true`, this does not break any functionality. I tested it by returning `false` for the basic `IGuiClickableArea` which is used for the vanilla plugin. The tooltip does not show up anymore and if you click, the recipes are still visible.

The use case:
I display a lot of information about the current recipe when hovering over the progress arrow in my custom screen. Since I don't want to break the general convention of opening the recipe page when clicking the progress arrow, I want this new API endpoint to avoid an overlapping tooltip.

Since the current implementation will also show the default tooltip when returning an empty list, I think a new default method is the best way. This could maybe changed in future versions as a breaking change because it doesn't make much sense that an empty list also uses the default tooltip. Why would you overwrite the method then?

I'd also appreciate a backport to 1.18.2 if possible.

### DIFF
--- a/Common/src/main/java/mezz/jei/common/gui/GuiEventHandler.java
+++ b/Common/src/main/java/mezz/jei/common/gui/GuiEventHandler.java
@@ -90,6 +90,7 @@ public class GuiEventHandler {
 			int guiLeft = screenHelper.getGuiLeft(guiContainer);
 			int guiTop = screenHelper.getGuiTop(guiContainer);
 			guiScreenHelper.getGuiClickableArea(guiContainer, mouseX - guiLeft, mouseY - guiTop)
+				.filter(IGuiClickableArea::isTooltipEnabled)
 				.map(IGuiClickableArea::getTooltipStrings)
 				.ifPresent(tooltipStrings -> {
 					if (tooltipStrings.isEmpty()) {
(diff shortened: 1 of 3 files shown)


mezz:
Thanks for contributing! Looks like you covered all the corner cases, looks good to me.

mezz:
Please feel free to open backport PRs, just make sure to start with the latest version like this one (JEI 11) then you can do 10, then 9, then 8, etc as far back as you like to go.

PR #2987, opened 2022-09-12
Prefer more abundant items first for recipe transfer

DBotThePony:
This change makes JEI provided recipe transferer smarter, by searching for most abundant ingredient(s) first.

This resolves cases of frustation where player quickly presses quickfill button, without checking *what* was filled in recipe slots, such as filling valuable resource (e.g. treated sticks) which is less abundant (in inventory) than common resource (e.g. regular sticks), and then crafting recipe, wasting valuable less-abundant resource.

New logic at deciding which ingredient to use is as follows:

- Search for all slots matching each ingredient
- Split all matched slots inside one ingredient into "equal" item groups (oak planks go to other oak planks, spruce to spruce, etc)
- Sort all result group lists to make groups with most items go first
- Sort all result group' sublists to make slots with least items go first
- Select first-match item from result stream

Demo:

https://user-images.githubusercontent.com/6339511/189647631-b94eeb48-0b9b-45da-8b8c-1de0fdef5354.mp4

https://user-images.githubusercontent.com/6339511/189647641-a914b8e9-2b1e-477d-a9a8-3b8b73cf5100.mp4

https://user-images.githubusercontent.com/6339511/189647642-4f33f3a8-100b-4853-a1f9-50246432f80b.mp4

https://user-images.githubusercontent.com/6339511/189647654-d6da4038-84e1-4cfb-ab9c-0bd7dabcbee1.mp4

### DIFF
--- a/Common/src/main/java/mezz/jei/common/transfer/BasicRecipeTransferHandlerServer.java
+++ b/Common/src/main/java/mezz/jei/common/transfer/BasicRecipeTransferHandlerServer.java
@@ -1,5 +1,6 @@
 package mezz.jei.common.transfer;
 
+import mezz.jei.core.util.Pair;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.inventory.AbstractContainerMenu;
 import net.minecraft.world.inventory.Slot;
@@ -38,7 +39,7 @@ public final class BasicRecipeTransferHandlerServer {
 			return;
 		}
 
-		Map<Slot, ItemStack> recipeSlotToRequiredItemStack = calculateRequiredStacks(transferOperations, player);
+		Map<Slot, Pair<Slot, ItemStack>> recipeSlotToRequiredItemStack = calculateRequiredStacks(transferOperations, player);
 		if (recipeSlotToRequiredItemStack == null) {
 			return;
 		}
@@ -132,8 +133,8 @@ public final class BasicRecipeTransferHandlerServer {
 	}
 
 	@Nullable
-	private static Map<Slot, ItemStack> calculateRequiredStacks(List<TransferOperation> transferOperations, Player player) {
-		Map<Slot, ItemStack> recipeSlotToRequired = new HashMap<>(transferOperations.size());
+	private static Map<Slot, Pair<Slot, ItemStack>> calculateRequiredStacks(List<TransferOperation> transferOperations, Player player) {
+		Map<Slot, Pair<Slot, ItemStack>> recipeSlotToRequired = new HashMap<>(transferOperations.size());
 		for (TransferOperation transferOperation : transferOperations) {
 			Slot recipeSlot = transferOperation.craftingSlot();
 			Slot inventorySlot = transferOperation.inventorySlot();
@@ -156,7 +157,7 @@ public final class BasicRecipeTransferHandlerServer {
 			}
 			ItemStack stack = slotStack.copy();
 			stack.setCount(1);
-			recipeSlotToRequired.put(recipeSlot, stack);
+			recipeSlotToRequired.put(recipeSlot, new Pair<>(inventorySlot, stack));
 		}
 		return recipeSlotToRequired;
 	}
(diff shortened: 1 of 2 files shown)


Common/src/main/java/mezz/jei/common/transfer/BasicRecipeTransferHandlerServer.java
```diff
@@ -38,7 +39,7 @@ public static void setItems(
 			return;
 		}
 
-		Map<Slot, ItemStack> recipeSlotToRequiredItemStack = calculateRequiredStacks(transferOperations, player);
+		Map<Slot, Pair<Slot, ItemStack>> recipeSlotToRequiredItemStack = calculateRequiredStacks(transferOperations, player);
```
mezz: Could you make this `Pair<Slot, ItemStack>` into a Record with named fields like `hint, required`, in order to clarify the code? Any code that accesses this pair's left or right gets pretty opaque.

Common/src/main/java/mezz/jei/common/transfer/BasicRecipeTransferHandlerServer.java
```diff
@@ -281,9 +283,14 @@ private static Set<Slot> merge(Map<Slot, ItemStack> result, Map<Slot, ItemStack>
 	}
 
 	@Nullable
-	private static Slot getSlotWithStack(Player player, ItemStack stack, List<Slot> craftingSlots, List<Slot> inventorySlots) {
+	private static Slot getSlotWithStack(Player player, ItemStack stack, List<Slot> craftingSlots, List<Slot> inventorySlots, Slot hint) {
 		Slot slot = getSlotWithStack(player, craftingSlots, stack);
+
 		if (slot == null) {
+			if (!hint.getItem().isEmpty() && ItemStack.isSameItemSameTags(stack, hint.getItem()) && hint.mayPickup(player)) {
```
mezz: just a style thing but I like to split up long lines like this so that there is one method call per line
```suggestion
                        Item hintItem = hint.getItem();
			if (!hintItem.isEmpty() && 
			    ItemStack.isSameItemSameTags(stack, hintItem) && 
			    hint.mayPickup(player)) {
```

Common/src/main/java/mezz/jei/common/transfer/RecipeTransferUtil.java
```diff
@@ -186,23 +192,117 @@ public static RecipeTransferOperationsResult getRecipeTransferOperations(
 	) {
 		RecipeTransferOperationsResult transferOperations = new RecipeTransferOperationsResult();
 
+		// Find groups of slots for each recipe input, so each ingredient knows list of slots it can take item from
+		// and also split them between "equal" groups
+		Map<IRecipeSlotView, Map<ItemStack, ArrayList<Map.Entry<Slot, ItemStack>>>> relevantSlots = new IdentityHashMap<>();
```
mezz: The type here is getting very long, can you refactor this into records or classes that are more descriptive?
I imagine `Map.Entry<Slot, ItemStack>` can be a record with named fields,
and something like `ArrayList<Map.Entry<Slot, ItemStack>>` could be encapsulated by a class that lets you manipulate it without exposing the underlying arraylist.

mezz (changes requested):
Thanks for contributing!
I think this needs to be slightly refactored for readability before I feel comfortable reviewing it for correctness. I left some notes that should help, and feel free to ask me any questions on Discord.

Common/src/main/java/mezz/jei/common/transfer/RecipeTransferUtil.java
DBotThePony: They are `Map.Entry`ies because i passthrough them directly from map, without copying them all over place, creating heap garbage. And `ArrayList<Map.Entry<Slot, ItemStack>>` is passed down to code below, where it is sorted.

mezz: The readability has been improved thanks!
I think you will find that any performance impact from the extra garbage generated here is imperceptible to players.
If we focus on optimizing as top priority, we wouldn't use Streams or Optional either, for example.

In general I think the only slowdowns I've see in JEI are from grossly bad algorithms or poorly chosen data structures, I've yet to see anything caused by extra object overhead.

mezz (approved):

mezz:
Looks good, thank you!

PR #2997, opened 2022-09-25
Improve performance of checking fuels on Fabric

Technici4n:
I noticed that initializing fuels was taking around 500 ms in my MI dev, and with the help of the profiler I found the culprit. :wink:

image


mezz:
Yikes, thanks!

PR #3052, opened 2022-11-22
log gui change on debug mode only

Jonathing:
Hello!

This is a bit of a small PR, but I think it is at least somewhat helpful. I use JEI in my development workspaces, and I noticed this message appearing a lot in my debug console (I mod with the `DEBUG` log level). The similar messages in this class are locked behind a config entry for "debug mode" while this one is not, and this single log message can flood my console incredibly easily (i.e. switching from chat to game). Anyways, cheers and thanks for your work on this mod! It's helped a ton for me so far.


mezz:
Thanks for the contribution!
Sorry I didn't see this earlier, I can merge it now.

Jonathing:
No problem, mate! Glad to be of service.

PR #3066, opened 2022-12-21
chore: Add Gui and Library maven publish to Jenkinsfile

klikli-dev:


mezz (approved):

mezz:
Thank you!

Jonathing:
If you don't mind me asking, what is the purpose of this? This seems like stuff that can fit into their respective API jars.

mezz:
Splitting up the jars allows me to separate the code and enforce barriers between different parts.
This way I can create the "GUI" code in a way where it only depends on JEI's API and some internal helper classes.

By doing that, JEI's GUI has the same limited access to JEI as other mods do, which makes sure the API is fairly complete and useful for other mods.

PR #3111, opened 2023-01-26
[1.19.3] JEI menus use creative TooltipFlags instead of non-creative ones to match with the Creative Menu.

bconlon1:
As of 1.19.3, a new type of TooltipFlag was added on to the pre-existing NORMAL and ADVANCED ones, gotten by using `TooltipFlag.Default#asCreative()` on either of these flags. This is used in the Creative menu as a way of having additional tooltip information for items displayed in tabs and not in the player's inventory.

This PR applies the change made to Creative menu tooltips to JEI's menus as well, due to the similar case of item display, which can be useful for mods that use this new way of determining what to display in a tooltip to have additional information tied to an item outside of the inventory.


mezz:
Sorry for the delayed response. This was added here: https://github.com/mezz/JustEnoughItems/commit/694752ac51b382eb4e2737acf28180d7bab7a16f

PR #3113, opened 2023-01-27
Add a 1.19 config for the position of new bookmarks

WaitingIdly:
This PR adds a new (advanced) config option to control if a new bookmark is prepended or appended. This also changes the default behavior to append.

I dont know what the lang controls in-game, but I followed the pattern of the other lang entries.

### DIFF
--- a/Common/src/main/resources/assets/jei/lang/en_us.json
+++ b/Common/src/main/resources/assets/jei/lang/en_us.json
@@ -137,7 +137,7 @@
   "config.jei.advanced.giveMode": "Give Mode",
   "config.jei.advanced.giveMode.comment": "Choose if JEI should give ingredients direct to the inventory (inventory) or pick them up with the mouse (mouse_pickup).",
   "config.jei.advanced.addBookmarksToFront": "Prepend New Bookmarks",
-  "config.jei.advanced.addBookmarksToFront.comment": "When enabled, adds new bookmarks to the front of the bookmark list.",
+  "config.jei.advanced.addBookmarksToFront.comment": "When true, adds new bookmarks to the front of the bookmark list. When false, adds new bookmarks to the end of the bookmark list",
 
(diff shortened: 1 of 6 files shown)


Gui/src/main/java/mezz/jei/gui/config/ClientConfig.java
```diff
@@ -42,6 +43,11 @@ public ClientConfig(IConfigSchemaBuilder schema) {
 			false,
 			"Enable cheating items into the hotbar by using the shift+number keys."
 		);
+		addBookmarksToFront = advanced.addBoolean(
+			"AddBookmarksToFrontEnabled",
+			false,
```
mezz:
```suggestion
			true,
```
Please keep the existing behavior here, users can configure it but I do not want to cause confusion by changing it by default.

mezz (changes requested):

Common/src/main/resources/assets/jei/lang/en_us.json
```diff
@@ -136,6 +136,8 @@
   "config.jei.advanced.maxRecipeGuiHeight.comment": "The maximum height of the recipe GUI.",
   "config.jei.advanced.giveMode": "Give Mode",
   "config.jei.advanced.giveMode.comment": "Choose if JEI should give ingredients direct to the inventory (inventory) or pick them up with the mouse (mouse_pickup).",
+  "config.jei.advanced.addBookmarksToFront": "Prepend New Bookmarks",
+  "config.jei.advanced.addBookmarksToFront.comment": "When enabled, adds new bookmarks to the front of the bookmark list.",
```
mezz:
```suggestion
  "config.jei.advanced.addBookmarksToFront.comment": "When true, adds new bookmarks to the front of the bookmark list. When false, adds new bookmarks to the end of the bookmark list",
```

Issue #1564, opened 2019-04-26
Change save location of Bookmarks.ini (bookmark save file)


LezChap:
There's debate in other requests on whether bookmarks should exist per world, or be account-wide.  The decision on that MAY influence where you move the file's save location (within the World/Save folder, or the main MC directory like JourneyMap or Prestige's data), but is beyond the scope of my actual request.

My request?  Move the Bookmarks.ini file's save location SOMEWHERE outside the config folder.  When updating modpacks, this can result in the bookmarks file being erased.  Even if updated manually into a new instance, finding a file in the config folder to move over to a new MultiMC instance is a bit counter-intuitive.

Thanks!

mezz:
Thanks for the report!
I think it makes sense to make it tied to the world.

LezChap:
I could see making it carry over between worlds make sense also...For example, I'm looking at building some multiblock or other complex structure in my survival world, but I want to test it first...I favorite everything I use in the build in my test world, then move back in the survival world to actually build it.

choekstr:
99.9999% of the people get their jei bookmarks overwritten by some useless items the packdev had in their config while testing, but that infinitesimally small amount of people that want to use the same bookmarks on a new world can copy over their config from another save folder. 

The way it is now, it's a permanent overwrite every pack update.
The proposed way would never overwrite and you can copy over from another world/save folder if needed. 

This should be a no brainer, why is there even a question on how this should be implemented?

mezz:
Bookmark configs will now be stored in a nested folder.
For example, in my local test world they are in:
`/config/jei/world/local/testworld/bookmarks.ini`

If you want to keep bookmarks across worlds, you can copy-paste them from one world folder to another.

This structure should prevent the configs from being overwritten by pack devs, as long as their world name or server info is different from yours.

Eric-01:
@mezz Couldn't JEI use the standard location where the vast majority of today's Forge mods save their world/server-specific configurations — at `~/.minecraft/saves/<the_world_name>/serverconfig` instead?

mezz:
> @mezz Couldn't JEI use the standard location where the vast majority of today's Forge mods save their world/server-specific configurations — at `~/.minecraft/saves/<the_world_name>/serverconfig` instead?

I thought that this directory was only intended for configs synced from the server to the client, but if it's more widely used then I can take a look and see if this can work for these configs.

mezz:
[...]

PR #3115, opened 2023-01-27
Move the 1.12.2 bookmark location out of the config folder

WaitingIdly:
something like this was suggested in #1564, but was settled via having bookmarks be per-world. Since 1.12.2 has settled on configs being per-instance, having it be directly in the instance folder (akin to journeymap) is a reasonable solution.
to make sure its purpose is understood, it was also renamed from `bookmarks.ini` to `jei_bookmarks.ini`.

this idea was first contributed to a similar project here: https://github.com/CleanroomMC/HadEnoughItems/pull/12

### DIFF
--- a/src/main/java/mezz/jei/config/Config.java
+++ b/src/main/java/mezz/jei/config/Config.java
@@ -303,11 +303,24 @@ public final class Config {
 			}
 		}
 
+		bookmarkFile = new File("./", "jei_bookmarks.ini");
+		File oldBookmarkFile = new File(jeiConfigurationDir, "bookmarks.ini");
+		if (!bookmarkFile.exists() && oldBookmarkFile.exists()) {
+			try {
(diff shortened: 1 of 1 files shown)


src/main/java/mezz/jei/config/Config.java
```diff
@@ -303,11 +303,24 @@ public static void preInit(FMLPreInitializationEvent event) {
 			}
 		}
 
+		bookmarkFile = new File("./", "jei_bookmarks.ini");
```
mezz:
```suggestion
        File minecraftDir = new File(Loader.instance().getConfigDir().getParent());
		bookmarkFile = new File(minecraftDir, "jei_bookmarks.ini");

```
I think we should be careful to get the right directory here, I am not sure what weird setups people might have with unusual `cwd`s.

mezz (changes requested):

PR #3159, opened 2023-03-20
Add JeiHelpers access to VanillaCategoryExtension

Edivad99:
I made these changes because I would like to draw an item in the gui when using the `IVanillaCategoryExtensionRegistration` interface.
I have attached an example of how I would like to do the GUI.

Screenshot 2023-03-20 alle 20 49 15

### DIFF
--- a/CommonApi/src/main/java/mezz/jei/api/registration/IVanillaCategoryExtensionRegistration.java
+++ b/CommonApi/src/main/java/mezz/jei/api/registration/IVanillaCategoryExtensionRegistration.java
@@ -16,6 +16,8 @@ import mezz.jei.api.recipe.category.extensions.vanilla.crafting.ICraftingCategor
 public interface IVanillaCategoryExtensionRegistration {
 	/**
 	 * {@link IJeiHelpers} provides helpers and tools for addon mods.
+	 *
+	 * @since 12.4.0
 	 */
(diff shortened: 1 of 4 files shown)


CommonApi/src/main/java/mezz/jei/api/registration/IVanillaCategoryExtensionRegistration.java
```diff
@@ -13,6 +14,10 @@
  * {@link IModPlugin#registerVanillaCategoryExtensions(IVanillaCategoryExtensionRegistration)}
  */
 public interface IVanillaCategoryExtensionRegistration {
+	/**
+	 * {@link IJeiHelpers} provides helpers and tools for addon mods.
+	 */
```
mezz: Can you please bump the `specificationVersion` number in `gradle.properties` to `12.4.0`
https://github.com/mezz/JustEnoughItems/blob/c6c115cfad74659c56cfbc293992e064f9deae0f/gradle.properties#L41

and write `@since 12.4.0` in the javadoc here so that people know which version of JEI to depend on if they need this feature?

mezz (changes requested):
Looks like a good idea to me, thank you for contributing!

mezz (approved):

mezz:
Thank you for contributing!

Issue #2953, opened 2022-08-11
[Suggestion]: Allow adding recipes to the bookmark


vfyjxf:
### Suggestion

When I add an ingredient to a bookmark, I often do so to see its recipe more quickly (because I can't remember it).
However, an ingredient usually has many recipes and I can't easily find the recipe I need. Therefore, allowing recipes to be added to a bookmark may be a good idea.

It should provide a feature that allows players to quickly view recipes (this allows players to distinguish between different recipes) and jump to marked recipes when viewing them through the recipe bookmark.
It may also need to provide a feature to quickly transfer recipes.

vfyjxf:
I've basically implemented this in my mod [Jei Utilities](https://github.com/vfyjxf/JEI-Utilities/tree/1.18.2), but it's so invasive and I'm not very good at rendering that it can often be problematic.
Here is a demo of it:

![dev(1)](https://user-images.githubusercontent.com/42764413/184156926-633f0b04-8891-48cd-ab84-ed22467d2de4.gif)

vfyjxf:
But what is inside Recipe Gui is not necessarily a recipe, and some mod authors like to add information or even a multi-block display inside. So it may still need some modification. (Because I determine whether to add recipes by identifying their output, they may not have an output)

mezz:
This is a cool feature.
It looks like @Buuz135 worked on something similar recently as well https://twitter.com/Buuz135mods/status/1555902527102672897
I'd be happy to integrate a polished version of this feature into JEI

mezz:
I am adding support here https://github.com/mezz/JustEnoughItems/pull/3574

PR #3164, opened 2023-03-26
Add Recipe Bookmark

vfyjxf:
It implements most of the features mentioned in https://github.com/mezz/JustEnoughItems/issues/2953 and makes some improvements, but it seems to have some unfinished code, but I've been busy with an exam recently so I probably won't be able to finish the rest of it, so I thought I'd submit it first up.


vfyjxf:
The unfinished code focuses on adding a button for RecipeLayout to bookmark recipes ，and another mode of RecipeBookmarkTooltip which counts the ingredients in a recipe and displays a simple summary on the tooltip.

PR #3178, opened 2023-04-06
make JEI available for 1.7.10 pls

WorthlessTrash:
Im stupid and am using old !@#$@# aka 1.7.10 and would really really appreciate it if me an idiot had the opportunity to know crafting recipes oh and for other ppl i guess.


mezz:
Sorry there is no way to do this, mods do not implement JEI plugins for 1.7.10. 
JEI for that version would only be able to support vanilla minecraft recipes, which is not very useful.

PR #3201, opened 2023-04-29
Add ability to return IRecipeManagerPlugin to the IJeiRuntime

skiprocks999:
Allows IJeiRuntimes to return their default handling of the IRecipeManagerPlugin

This allows for more general custom handling for recipe lookups. For example, you can hit 'R' on an item that contains a fluid, get the fluid from the item, and then return the result of looking up the fluid.

To prevent forks of JEI from breaking, the implementation in the interface returns null by default

I did this in notepad for reference, so apologies if the spaces are kinda borked


mezz:
Thanks for the suggestion!
I think this would work more cleanly as a config option in JEI, so that users can turn it on and off, instead of having it added mysteriously by another mod. I have added it here: 76278d489072fc1d3e3477df163228ae03e6e42e

PR #3210, opened 2023-05-08
Add JeiRegistrationDelegate API

exaskye:
Adds an advanced API to provide a stand-between when JEI is calling methods on the registered plugins. Only one registration delegate may exist. By default, it simply calls the method with no further logic — equivalent to current behavior.

This is useful in certain cases to override other plugins, especially JEI's built-in vanilla plugins. It also makes the JEI loading process more inspectable and debuggable without code modification.

I confirmed this builds and runs for Fabric, but I cannot build Forge due to the following error:

```
> Task :Forge:compileTestJava FAILED
/home/una/Source/JustEnoughItems/Forge/src/test/java/mezz/jei/test/lib/TestClientConfig.java:9: error:
  TestClientConfig is not abstract and does not override abstract method isLookupFluidContents() in IClientConfig

public class TestClientConfig implements IClientConfig {
       ^
1 error

FAILURE: Build failed with an exception.
```

### DIFF
--- a/CommonApi/src/main/java/mezz/jei/api/registration/JeiRegistrationDelegate.java
+++ b/CommonApi/src/main/java/mezz/jei/api/registration/JeiRegistrationDelegate.java
@@ -1,0 +1,29 @@
+package mezz.jei.api.registration;
+
+import mezz.jei.api.IModPlugin;
+
+import java.util.function.Consumer;
+
+/**
+ * Advanced API to provide a stand-between when JEI is calling methods on the registered plugins.
+ * Only one registration delegate may exist. By default, it simply calls the method with no further
+ * logic.
+ */
+public interface JeiRegistrationDelegate {
+
+	static void set(JeiRegistrationDelegate delegate) {
+		JeiRegistrationDelegateHolder.current = delegate;
+	}
+
+	static JeiRegistrationDelegate get() {
+		return JeiRegistrationDelegateHolder.current;
+	}
+
+	void callOnPlugin(JeiRegistrationStep step, IModPlugin plugin, Consumer<IModPlugin> func);
+
+}
+
+// Can't have a private in an interface, including an inner class. So this will do.
+class JeiRegistrationDelegateHolder {
+	static JeiRegistrationDelegate current = (step, plugin, func) -> func.accept(plugin);
+}
--- a/CommonApi/src/main/java/mezz/jei/api/registration/JeiRegistrationStep.java
+++ b/CommonApi/src/main/java/mezz/jei/api/registration/JeiRegistrationStep.java
@@ -1,0 +1,33 @@
+package mezz.jei.api.registration;
+
+/**
+ * The order of values in this enum is not guaranteed. Use {@code ordinal()} with caution, if ever.
+ * @see JeiRegistrationDelegate
+ */
+public enum JeiRegistrationStep {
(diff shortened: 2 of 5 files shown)


mezz:
Thanks for contributing!
I have fixed the broken Forge test, it should work now if you rebase.

What is the specific motivation for this feature? You mention that this will make it easier to inspect JEI and debug it, but this feature could actually make things quite hard to debug if anyone uses it. I think there should be a relatively high bar to adding complexity to the system this way, so the more you can explain your use case the better I can justify it.

exaskye:
This API offers an explicit way to hand off control to another mod to suppress JEI GUI and allow other mods to utilize the registerRuntime endpoint to take control of the frontend, the extra flexibility also allows the mod to reduce redundancy if otherwise handled (i.e. custom implementation of vanilla recipes).

mezz:
Ok thank you, I understand what this is intended for now.

CommonApi/src/main/java/mezz/jei/api/registration/JeiRegistrationDelegate.java
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	void callOnPlugin(JeiRegistrationStep step, IModPlugin plugin, Consumer<IModPlugin> func);
+
+}
+
+// Can't have a private in an interface, including an inner class. So this will do.
+class JeiRegistrationDelegateHolder {
+	static JeiRegistrationDelegate current = (step, plugin, func) -> func.accept(plugin);
+}
```
mezz, line 29: as a design rule, I do not store global variables like this in the API

I think this could be implemented a little more cleanly if we did something like `IModPlugin`, where JEI can discover  them and use the non-JEI one if found (and crash or pick the first one if there are multiple).

We could create two new types: 

1. `IJeiRuntimePlugin` which has
`void registerRuntime(IRuntimeRegistration registration)`
and deprecate that method in `IModPlugin`

2. `IJeiPluginLoaderPlugin` (lol?) which has
`void callOnPlugin(JeiRegistrationStep step, IModPlugin plugin, Consumer<IModPlugin> func)`

CommonApi/src/main/java/mezz/jei/api/registration/JeiRegistrationStep.java
```diff
@@ -0,0 +1,33 @@
+package mezz.jei.api.registration;
+
+/**
+ * The order of values in this enum is not guaranteed. Use {@code ordinal()} with caution, if ever.
+ * @see JeiRegistrationDelegate
+ */
```
mezz, line 6: all new stuff should have `@since #.#.#` in the javadoc. you can bump the version number in the gradle properties

mezz (changes requested):
This looks good to me! I'd just like to take a moment to use the plugin discovery system to make this cleaner. I have explained more in the comments

mezz:
The requests I made are kind of a pain so I'm going to see if I can work on it a bit

exaskye:
Yeah, I don't really know enough about the codebase to do half of what was asked, and I've been pretty busy as of late.

exaskye:
EMI has shipped its JEI support via mixins, and this PR has been dormant for a while with rejections I can't resolve, so I'm going to go ahead and close it.

PR #3218, opened 2023-05-15
[1.12] Fix duplication glitch when shift-clicking crafting recipes

TechLord22:
This PR fixes a duplication glitch when shift-clicking the "+" button to import recipes into the crafting table. 

When clearing the crafting grid, there is a scenario where not all of the items can fit into the player's inventory. JEI currently drops the items on the ground when this occurs. In doing so, it does not account for any items that were added to the player's inventory beforehand, so it drops the entire stack present in the grid. As a result, extra items are given to the player, and then the original stack is dropped on the ground.

This PR resolves this issue by subtracting the inserted amount before dropping the stack.

Reported issue where this bug was initially discovered: https://github.com/GregTechCEu/GregTech/issues/1775.


mezz:
Thank you for the fix!

PR #3219, opened 2023-05-15
WIP async loading

mezz:
This adds two config flags that are disabled by default:
1. enables async loading of JEI, which means it will not block the world load and it will continue loading in the background for a while before it's available to use.
2. enables parallel plugin loading. this is more of an experimental config option that will break stuff. Plugins can opt into this more intelligently by implementing the new `IAsyncModPlugin` instead of `IModPlugin`.


vfyjxf:
I think we could start by offering the async plugin as an API and encourage modder to use it;
only a few plugins would need to sync

PR #3220, opened 2023-05-17
Fix mismatch between display names

HwakYuk:
`IngredientManager.removeIngredientsAtRuntime` is broken for items with chat formatting codes. The problem is when adding an ingredient `IngredientInformationUtil.getDisplayName` is used which strips chat formatting after `ingredientHelper.getDisplayName`, while when removing only `ingredientHelper.getDisplayName` is used, so `ItemFilter` couldn't find any matching ingredient to delete. For example, this phenomenon occurs for AllTheCompressed mod's atm star blocks and nether star blocks.


mezz:
Thank you for the fix!

PR #3229, opened 2023-05-29
Initial Port to 1.20

shartte:
Does include smithing recipe category now, although it's unclear to me how the trim recipes actually work and if simply cycling through the inputs is sufficient. (I don't think the trim material is reflected correctly in the output, it might need special consideration).


shartte:
@mezz This should be ready to go now

PR #3235, opened 2023-06-04
Only rebuild ingredient filter when client resources reload

embeddedt:
Currently, when client resource packs are reloaded, JEI completely restarts itself. However the startup sequence is quite slow, and this can add 30-40 seconds to resource reloading in large modpacks.

After some thought I realized that most of this work should be unnecessary. The only data that can change when client resources reload are language strings and textures. Textures are already handled by the item renderer, so the only part of JEI that actually needs to be reloaded is the ingredient filter.

This PR changes the reload listener to call a new `rebuildItemFilter` method which simply saves the current list of ingredients, sorts them, and then recreates the search structure and reinserts the same list of ingredients. I have not directly tested it but this should also be completely compatible with mods that remove items in `onRuntimeAvailable`, as we simply reuse the current list at the point that the resource reload happened.

Scenarios tested:

* Running `/reload` on Forge fully reloads JEI, as normal.
* Reloading client resources via F3+T only reloads search structure.
* Verified that search structure is indeed reloaded by changing the language to Pirate Speak and searching for "timber" in JEI. ;)

This change is currently not implemented on Fabric as JEI doesn't seem to react to client resource reloading at all. Not sure if that is a bug or not. In any case, the only necessary change to enable this on Fabric would be to duplicate the change to `StartEventObserver`.

### DIFF
--- a/CommonApi/src/main/java/mezz/jei/api/runtime/IIngredientFilter.java
+++ b/CommonApi/src/main/java/mezz/jei/api/runtime/IIngredientFilter.java
@@ -45,4 +45,9 @@ public interface IIngredientFilter {
 	 * to get all the ingredients known to JEI, not just ones currently shown by the filter
 	 */
 	<T> List<T> getFilteredIngredients(IIngredientType<T> ingredientType);
+
+	/**
+	 * Rebuild the ingredient filter data structures. Called by JEI when client-side resources reload.
+	 */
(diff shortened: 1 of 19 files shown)


mezz:
Thanks for contributing!
I have updated this to work more internally, without API changes. I also added support for Fabric (like you pointed out) and for NeoForge.

Issue #2677, opened 2022-01-13
[1.18.1-9.2.1.71] "Gave 0 [Air]" message when clicking to retrieve item with full stack already in hand


sdarkwell:
**Hello,**

While in "Mouse Pickup" Cheat Mode, when clicking on an item with a full stack of said item already in hand, the console message "Gave 0 [Air]" is displayed in the chat.

My assumption is that it would be preferred for that message to be suppressed.

**Thank you and be well!**
\- S.

mezz:
Thanks for the report!
This issue has been fixed in all supported JEI versions.
Commit 212f9c745 added the correct `giveCount > 0` guard. The first release containing it was JEI 9.4.2 for Minecraft 1.18.2.

PR #3246, opened 2023-06-13
[1.18.1] Suppressed "Gave 0 [Air]" message when trying to obtain item with full stack of the item already in cursor.

cjapar2:
Fixes issue #2677.

Added the following if-statement to check if the current stack on the cursor is less than the maximum stack allowed. This makes it so that the game doesn't unnecessarily try to give you more of the item than it can. As a result, the "Give Air" message does not appear in the chat and console log.
```
if (player.inventoryMenu.getCarried().getCount() < player.inventoryMenu.getCarried().getMaxStackSize()) {
	Network.sendPacketToServer(packet);
}
```


mezz:
Thanks for contributing!
Your fix seems to target `setHotbarStack`, is that really what fixes the bug described in the issue?

Also if you want to fix this in 1.18.2, the normal branch to use is `1.18`
I just have the 1.18.1 branch for that specific version but it's not used by many people.

PR #3250, opened 2023-06-13
Github Actions Workflow

shartte:
Just runs a build on every push and bundles up the artifacts.

Example Run: https://github.com/shartte/JustEnoughItems/actions/runs/5260526283


mezz:
Thanks!

PR #3255, opened 2023-06-18
Recipe category decorators

rlnt:
## PR Information

This pull request implements the ability to attach decorators to recipe categories in order to draw additional elements, add tooltips, or even alter the tooltips of the category.

A decorator must implement `IRecipeCategoryDecorator` and has to be registered inside the `IAdvancedRegistration`. With different overloads, you have the option to register the extension for recipe categories, ~~recipe classes~~, and recipe types.
~~Furthermore, all recipe categories were exposed as an immutable list inside the `IJeiHelpers`. This allows registering extensions for categories of other mods or even globally for every recipe possible without the need to know the recipe type or depending on the mod providing the category.~~

## PR Merging Notes

~~Since I am not exactly sure how you handle the versioning, I marked every modification as `@since 15.1.0` because the last released 1.20 version was `v15.0.0`.~~

~~Also, some interfaces needed additions that include breaking changes.~~

~~The `IRecipeLayoutDrawable` now has the `getRecipeCategoryExtensions` method. If a mod implements a custom recipe layout drawable and they don't implement the new method, they would get a crash.~~

~~However, I could implement a default implementation where an empty collection is returned. This would fix the crash but the extensions will not work for custom implementations if modders don't override this behavior. It's up to you how to handle that.~~

~~The second breaking interface change is the `IRecipeManager` which now features a method to create an `IRecipeLayoutDrawable` with extensions.~~

## Example

A global extension that renders text when the recipe registry name includes the word `"minecraft"`.
```java
@Override
public void registerAdvanced(IAdvancedRegistration registration) {
    var recipeTypes = registration.getJeiHelpers().getAllRecipeTypes();
    recipeTypes.forEach(rt -> registration.addRecipeCategoryDecorator(rt, new Decorator<>()));
}

private static class Decorator<T> implements IRecipeCategoryDecorator<T> {
	@Override
	public void draw(
		T recipe, IRecipeCategory<T> recipeCategory, IRecipeSlotsView recipeSlotsView, GuiGraphics guiGraphics,
		double mouseX, double mouseY
	) {
		var id = recipeCategory.getRegistryName(recipe);
		if (id == null) return;

		if (id.toString().contains("minecraft")) {
			var posX = recipeCategory.getWidth() / 2;
			var posY = recipeCategory.getHeight() / 2;
			guiGraphics.drawCenteredString(Minecraft.getInstance().font, "Hello from Extension", posX, posY, 0xFF_FFFF);
		}
	}
}
```

Since the advanced registry also allows registering extensions for recipe types and you expose the vanilla recipe types inside the API, it's also possible to attach the extension to a specific vanilla category only.
```java
registration.addRecipeCategoryDecorator(RecipeTypes.ANVIL, new IRecipeCategoryDecorator<>() {
    @Override
    public void draw(
        IJeiAnvilRecipe recipe, IRecipeCategory<IJeiAnvilRecipe> recipeCategory,
        IRecipeSlotsView recipeSlotsView, GuiGraphics guiGraphics, double mouseX, double mouseY
    ) {
        // do something with anvil recipe here
    }
});
```

Of course, it's also possible to move the text in front of the items. But this image is supposed to show that the `PoseStack` isn't altered for extensions so you can transform it however you'd like.


mezz:
`IRecipeLayoutDrawable` and `IRecipeManager` are only implemented by JEI, so adding methods to these is not considered breaking changes for mods. Changing or removing methods is a breaking change though

mezz:
For the version, you can bump it to 15.1.0 in the `gradle.properties` file

CommonApi/src/main/java/mezz/jei/api/helpers/IJeiHelpers.java
```diff
@@ -61,6 +63,13 @@ public interface IJeiHelpers {
 	 */
 	Optional<RecipeType<?>> getRecipeType(ResourceLocation uid);
 
+	/**
+	 * Get all registered recipe categories.
+	 *
+	 * @since 15.1.0
+	 */
+	Collection<IRecipeCategory<?>> getRecipeCategories();
```
mezz: You can do this with the API without adding a new method,
```java
// IRecipeManager recipeManager;

recipeManager.createRecipeCategoryLookup()
    .get()
    .toList()
```

CommonApi/src/main/java/mezz/jei/api/recipe/category/extensions/IGlobalRecipeCategoryExtension.java
```diff
@@ -0,0 +1,67 @@
...
+import mezz.jei.api.gui.drawable.IDrawable;
+import mezz.jei.api.gui.ingredient.IRecipeSlotsView;
+import mezz.jei.api.helpers.IGuiHelper;
+import mezz.jei.api.recipe.category.IRecipeCategory;
+import net.minecraft.client.gui.GuiGraphics;
+import net.minecraft.network.chat.Component;
+
+import java.util.List;
+
+public interface IGlobalRecipeCategoryExtension<T> {
```
mezz: This interface should have a doc to explain its intended purpose, and an `@since 15.1.0` so people know what minimum version of JEI to set for their mod.

CommonApi/src/main/java/mezz/jei/api/recipe/category/extensions/IGlobalRecipeCategoryExtension.java
```diff
@@ -0,0 +1,67 @@
...
+	 *
+	 * @param tooltips        the existing tooltip strings.
+	 * @param recipe          the current recipe being drawn.
+	 * @param recipeCategory  the recipe category of the recipe.
+	 * @param recipeSlotsView a view of the current recipe slots being drawn.
+	 * @param mouseX          the X position of the mouse, relative to the recipe.
+	 * @param mouseY          the Y position of the mouse, relative to the recipe.
+	 * @return tooltip strings. If there is no tooltip at this position, return an empty list.
+	 */
+	default List<Component> decorateExistingTooltips(List<Component> tooltips, T recipe, IRecipeCategory<T> recipeCategory, IRecipeSlotsView recipeSlotsView, double mouseX, double mouseY) {
```
mezz: I do not see much utility in having this method *and* `getTooltipStrings`.
Could everything be done in this method, and we can remove `getTooltipStrings`?

CommonApi/src/main/java/mezz/jei/api/registration/IAdvancedRegistration.java
```diff
@@ -17,4 +20,25 @@ public interface IAdvancedRegistration {
...
+	 * @since 15.1.0
+	 */
+	<T> void addGlobalRecipeCategoryExtension(IRecipeCategory<T> recipeCategory, IGlobalRecipeCategoryExtension<T> extension);
+
+	/**
+	 * Register a {@link IGlobalRecipeCategoryExtension} for a recipe class.
+	 *
+	 * @since 15.1.0
+	 */
+	<T> void addGlobalRecipeCategoryExtension(Class<T> recipeClass, IGlobalRecipeCategoryExtension<T> extension);
```
mezz: I want to avoid using `Class` wherever possible, the `RecipeType<T>` is a lot safer from a validation perspective. It's harder to make a mistake calling a method where the compiler wants a `RecipeType<T>`.
We should remove this method.

Library/src/main/java/mezz/jei/library/load/registration/AdvancedRegistration.java
```diff
@@ -29,6 +36,43 @@ public void addRecipeManagerPlugin(IRecipeManagerPlugin recipeManagerPlugin) {
...
+		}
+    }
+
+	@SuppressWarnings({"rawtypes", "unchecked"})
+	@Override
+	public <T> void addGlobalRecipeCategoryExtension(RecipeType<T> recipeType, IGlobalRecipeCategoryExtension<T> extension) {
+		ErrorUtil.checkNotNull(recipeType, "recipeType");
+		ErrorUtil.checkNotNull(extension, "extension");
+
+		for (IRecipeCategory<?> recipeCategory : jeiHelpers.getRecipeCategories()) {
```
mezz: instead of `jeiHelpers.getRecipeCategories()`

you can create a stream with
```java
recipeManager.createRecipeCategoryLookup()
    .limitTypes(List.of(recipeType))
    .get()
```

CommonApi/src/main/java/mezz/jei/api/recipe/category/extensions/IGlobalRecipeCategoryExtension.java
mezz: It seems a bit dangerous to be able to register one of these for "any recipe category that handles some recipe type".
If we have unknown categories, should we draw on them? I think we might not.
To make this safer, could it be something like `<T, R extends IRecipeCategory<T>>`?

mezz: This name is pretty long and generic, we should come up with a better name for this and its methods that help make it feel like a more concrete concept.
I'll try to help think of something.

mezz (changes requested):

CommonApi/src/main/java/mezz/jei/api/recipe/category/extensions/IGlobalRecipeCategoryExtension.java
rlnt: Sure but I thought it might be hard to notice the behavior when you have to pass the original list back. If some modders return their own list without the already existing elements of the category, they will essentially delete them.
I'll merge the two methods and will make sure to properly mention it in the doc.

CommonApi/src/main/java/mezz/jei/api/recipe/category/extensions/IGlobalRecipeCategoryExtension.java
rlnt: I agree. Sadly, `IRecipeCategoryExtension` is already taken and I couldn't come up with another name.
What about something like `ICategoryDecorator`?

Library/src/main/java/mezz/jei/library/load/registration/AdvancedRegistration.java
rlnt: And how would I grab an instance of the recipe manager at this point?

rlnt: The only option I see to do it in the way you mentioned is by storing two fields inside the advanced registration. One map for RecipeCategory -> Extensions and one for RecipeType -> Extensions and then resolve them in the RecipeGuiLogic.

CommonApi/src/main/java/mezz/jei/api/recipe/category/extensions/IGlobalRecipeCategoryExtension.java
mezz: What is the use-case for altering the existing tooltip?
Should this interface be used to optionally override the old tooltip, without passing it in?

CommonApi/src/main/java/mezz/jei/api/recipe/category/extensions/IGlobalRecipeCategoryExtension.java
LLytho: I don't really get why it's dangerous to be able to register them, as the user is not able to register an invalid extension to a category anyways. 
What do you mean by unknown category? The extensions can be registered to any kind of category the user wants, so I think we should allow drawing on them.

Library/src/main/java/mezz/jei/library/load/registration/AdvancedRegistration.java
mezz: Ah ok I see the problem.
You can pass the list of recipe categories into `AdvancedRegistration` here:
https://github.com/mezz/JustEnoughItems/blob/2bbf6670b43d5d1221f286e0d13f4813f2df1a2b/Library/src/main/java/mezz/jei/library/load/PluginLoader.java#L118-L124

and avoid adding the extra access to it in the API

CommonApi/src/main/java/mezz/jei/api/recipe/category/extensions/IGlobalRecipeCategoryExtension.java
mezz: `IRecipeCategoryDecorator` sounds good to me

CommonApi/src/main/java/mezz/jei/api/recipe/category/extensions/IGlobalRecipeCategoryExtension.java
rlnt: You have three opportunities with the decorate method:
1. Remove all original tooltips of the category this extension is registered to by always returning an empty list or only with your custom tooltips.
2. Change existing tooltips of the category by modifying the elements in the list.
3. Add new custom tooltips by appending them to the list.

The method is able to basically do anything with the tooltips of the category it's bound to. This makes the `getTooltipStrings` method obsolete, that's why I decided to remove it.
Each of the things mentioned above has its own use cases. For example, staging mods could disable all tooltips until the player has the stage for the recipe. Or the staging mod draws an icon/text for the required stage on the category and adds a tooltip to it.

CommonApi/src/main/java/mezz/jei/api/recipe/category/extensions/IGlobalRecipeCategoryExtension.java
mezz: ~~Here's a scenario I have in mind:~~
~~Imagine I want to draw something on top of a Thaumcraft recipe where I want to draw a blue circle in the center of the recipe between some ingredients.~~
~~I create an extension here and not knowing the difference, I register it for the RecipeType.~~
~~I do not know that there is an addon to Thaumcraft that creates a recipe category for this same RecipeType but it is drawn very differently.~~
~~Now my extension is drawing on top of that addon's recipes in a way that is broken.~~

mezz: Hmm actually I might not know how JEI works.
There can only be one recipe category per recipe type:
https://github.com/mezz/JustEnoughItems/blob/8a65df7e56a1c00d4dd4153260902e4f26a7401a/CommonApi/src/main/java/mezz/jei/api/recipe/RecipeType.java#L9

mezz: This is fine 👍

CommonApi/src/main/java/mezz/jei/api/recipe/IRecipeManager.java
```diff
@@ -111,13 +113,15 @@ public interface IRecipeManager {
...
+	 * @param decorators	 the decorators for the recipe category.
 	 * @param recipe         the specific recipe to draw.
 	 * @param focusGroup     the focuses of the recipe layout.
 	 *
-	 * @since 11.5.0
+	 * @since 15.1.0
 	 */
 	<T> Optional<IRecipeLayoutDrawable<T>> createRecipeLayoutDrawable(
 		IRecipeCategory<T> recipeCategory,
+		Collection<IRecipeCategoryDecorator<T>> decorators,
```
mezz: I think we can avoid changing this and avoid exposing `getRecipeCategoryDecorators`, they can be applied internally all the time.

CommonApi/src/main/java/mezz/jei/api/registration/IAdvancedRegistration.java
```diff
@@ -17,4 +20,18 @@ public interface IAdvancedRegistration {
 	 * Register your own {@link IRecipeManagerPlugin} here.
 	 */
 	void addRecipeManagerPlugin(IRecipeManagerPlugin recipeManagerPlugin);
+
+	/**
+	 * Register a {@link IRecipeCategoryDecorator} for a recipe category.
+	 *
+	 * @since 15.1.0
+	 */
+	<T> void addRecipeCategoryDecorator(IRecipeCategory<T> recipeCategory, IRecipeCategoryDecorator<T> decorator);
```
mezz: This is the same as calling
`addRecipeCategoryDecorator(recipeCategory.getRecipeType(), decorator)`
so I think we can remove it

CommonApi/src/main/java/mezz/jei/api/recipe/category/extensions/IGlobalRecipeCategoryExtension.java
rlnt: > Hmm actually I might not know how JEI works.

This made my day 😅

CommonApi/src/main/java/mezz/jei/api/registration/IAdvancedRegistration.java
rlnt: It was added because adding the decorator to a recipe type requires iteration of all recipe categories in order to validate the type and whether it has a category attached to it.
It may not be that impactful. I'll remove it.

CommonApi/src/main/java/mezz/jei/api/recipe/IRecipeManager.java
rlnt: Not exactly sure what you mean by that.

CommonApi/src/main/java/mezz/jei/api/registration/IAdvancedRegistration.java
mezz: Yeah there are not that many recipe categories to iterate over so it's not a big deal, maybe 1000s in a big pack.
For optimal performance we could use a `Map<RecipeType, RecipeCategory>` internally, I'm sure I already create that somewhere else

CommonApi/src/main/java/mezz/jei/api/recipe/IRecipeManager.java
mezz: I think we can implement this logic internally without making any changes to `IRecipeManager` in the api

CommonApi/src/main/java/mezz/jei/api/gui/IRecipeLayoutDrawable.java
```diff
@@ -87,6 +89,12 @@ default Optional<ItemStack> getItemStackUnderMouse(int mouseX, int mouseY) {
 	 */
 	IRecipeCategory<R> getRecipeCategory();
 
+	/**
+	 * Get the decorators for the recipe category of this recipe layout.
+	 * @since 15.1.0
+	 */
+	Collection<IRecipeCategoryDecorator<R>> getRecipeCategoryDecorators();
```
mezz: I think we can remove this and keep it as an internal detail

CommonApi/src/main/java/mezz/jei/api/recipe/IRecipeManager.java
rlnt: Since I am not sure what you have in mind for that, feel free to add it to the branch yourself. Maybe together with the `Map<RecipeType, RecipeCategory>` idea from the other review. :)

CommonApi/src/main/java/mezz/jei/api/gui/IRecipeLayoutDrawable.java
mezz: One thing to keep in mind is that anything added to the API will stay around for a long time. Any changes we make will need to go through a version-long deprecation cycle. I try to add the minimum required for a new feature, we can decide to add more later if needed.

CommonApi/src/main/java/mezz/jei/api/recipe/IRecipeManager.java
mezz: Sounds good. Can you give me push access to this branch?
I think you have to select "Allow edits from maintainers" on the PR

rlnt: That usually is enabled by default. I can't seem to find a setting where you can set this after the PR is open already. This may be due to the fact that we forked it into our organization.
I'll just add it as an outside collaborator for the forked repo as an alternative solution.

mezz: Thanks, that worked. I have pushed my changes, please let me know if this still looks good to you.

rlnt:
Your changes look good to me! 👍

However, I am seeing one more problem in general.
When you want to register a decorator for all categories or for other categories you don't have access to by grabbing them from the global list, it is currently not possible to grab the existing categories inside the `IModPlugin#registerAdvanced` method.

My test plugin class looks like this:
```java
@JeiPlugin
public class TestPlugin implements IModPlugin {

    private IJeiRuntime runtime;

    @Override
    public ResourceLocation getPluginUid() {
        return new ResourceLocation("jei", "testplugin");
    }

    @Override
    public void onRuntimeAvailable(IJeiRuntime jeiRuntime) {
        this.runtime = jeiRuntime;
    }

    @Override
    public void registerAdvanced(IAdvancedRegistration registration) {
        if (runtime == null) return;

        var categories = runtime.getRecipeManager().createRecipeCategoryLookup().get();
        categories.forEach(c -> registration.addRecipeCategoryDecorator(c.getRecipeType(), new Extension<>()));
    }

    private static class Extension<T> implements IRecipeCategoryDecorator<T> {
        @Override
        public void draw(
            T recipe, IRecipeCategory<T> recipeCategory, IRecipeSlotsView recipeSlotsView, GuiGraphics guiGraphics,
            double mouseX, double mouseY
        ) {
            var id = recipeCategory.getRegistryName(recipe);
            if (id == null) return;

            if (id.toString().contains("minecraft")) {
                var posX = recipeCategory.getWidth() / 2;
                var posY = recipeCategory.getHeight() / 2;
                guiGraphics.drawCenteredString(Minecraft.getInstance().font, "Hello from Extension", posX, posY, 0xFF_FFFF);
            }
        }
    }
}
```

`IModPlugin#onRuntimeAvailable` is called after all registry calls have been made to the plugins so at the point in time where the decorators need to be registered, it's not possible to grab the categories from the runtime. Is there another way to access the list or the recipe manager to register decorators for other categories?

It only works after a reload because I store the runtime.

mezz:
To get a specific recipe type, you can use
`mezz.jei.api.helpers.IJeiHelpers#getRecipeType`

I can add a new method to get all of them,
`mezz.jei.api.helpers.IJeiHelpers#getAllRecipeTypes`

mezz:
ok @DAmNRelentless I have added `IJeiHelpers#getAllRecipeTypes` and added a debug category decorator to the PR, let me know how this looks to you.

rlnt:
lgtm
Thank you for your responsiveness and fast work. I am pretty happy with the solution. 💚

I updated my initial PR description to show examples of how to implement the new feature in case someone needs a reference when stumbling across the new API endpoints.

PR #3263, opened 2023-06-30
[1.20] Add / Fix icon paths in mods.toml and fabric.mod.json

HyCraftHD:


mezz:
Thank you for the fix!

PR #3306, opened 2023-08-17
fix items dupe

Sleepwalkerx:
If the inventory can take an item and is overflowing items will be dupe


sennhann29 (approved):

mezz:
Thank you for the fix!

PR #3353, opened 2023-11-10
support use short modid to search mod ingredients

vfyjxf:
Preview: 
image
image
The rule is very simple. Instead of a modid, the player may use a string consisting of the first two letters of each word of the mod name, or the first char of each word if the mod name contains only two words.


mezz:
Thank you for contributing! I think it's a good idea.
I have made this a bit more configurable and moved things to a new registration class.

PR #3385, opened 2024-01-01
Fix keybinds not being configurable for Fabric

KosmosPrime:
Fixes that keybindings are not displayed in the Minecraft controls menu for Fabric and can thus not be rebound, but at the cost of losing modifiers (which are not supported by plain Fabric). Contains optional support for [AMECS](https://modrinth.com/mod/amecs) to automatically re-enable key modifiers.


mezz:
Thank you for contributing!

PR #3398, opened 2024-01-13
Initial 1.20.4 port

malte0811:
Status of the port:
- NeoForge: Working (though not thoroughly tested)
- Fabric: Same
- Forge: Would probably work if gradle didn't produce cryptic errors on both `build` and `runClient`
- Code still needs some cleanup

This is based on #3363 by @shartte.

### DIFF
--- a/NeoForgeApi/src/main/java/mezz/jei/api/neoforge/NeoForgeTypes.java
+++ b/NeoForgeApi/src/main/java/mezz/jei/api/neoforge/NeoForgeTypes.java
@@ -1,0 +1,35 @@
+package mezz.jei.api.neoforge;
+
+import mezz.jei.api.ingredients.IIngredientType;
+import mezz.jei.api.ingredients.IIngredientTypeWithSubtypes;
+import net.minecraft.world.level.material.Fluid;
+import net.neoforged.neoforge.fluids.FluidStack;
+
+/**
+ * Built-in {@link IIngredientType} for NeoForge Minecraft.
+ *
+ * @since 9.6.0
+ */
+public final class NeoForgeTypes {
+    /**
+     * @since 9.7.0
+     */
+    public static final IIngredientTypeWithSubtypes<Fluid, FluidStack> FLUID_STACK = new IIngredientTypeWithSubtypes<>() {
+        @Override
+        public Class<? extends FluidStack> getIngredientClass() {
+            return FluidStack.class;
+        }
+
+        @Override
+        public Class<? extends Fluid> getIngredientBaseClass() {
+            return Fluid.class;
+        }
+
+        @Override
+        public Fluid getBase(FluidStack ingredient) {
+            return ingredient.getFluid();
+        }
+    };
+
+    private NeoForgeTypes() {}
+}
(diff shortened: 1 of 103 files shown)


Spinoscythe:
can I add more commits to this PR?, I have free time

malte0811:
I think this is ready for review now. Known issues:
- The NF subprojects are mostly copies of the MCF subprojects with updated imports. I do not think there is much to be done about this, and since they are going to diverge further in the future it is probably a good idea anyway.
- Decompiling does not work for the Forge subprojects. I do not think this can be fixed without breaking decompilation for NeoForge. Since NF does not support binpatching in dev at all, that option would make even building for NF impossible, so I decided to keep MCF decompilation broken instead. The current status means that anything triggering source patching in FG fails, which unfortunately includes IDE import with the `Forge`/`ForgeApi` projects enabled.  
Technical details: Both NG and FG load a version of `DiffPatch` (`codechicken:DiffPatch:1.5.0.29` vs `net.minecraftforge:DiffPatch:2.0.12`). Both use the same package names, even though the maven artifact names suggest otherwise. So effectively one of these ends up getting prioritized over the other. Neither of the possible priority orders works:
  - If the version requested by NG is used (current state), FG fails during source patching because `PatchOperation$Builder#verbose(boolean)` does not exist in that version.
  - If the version requested by FG is used (by "preloading" FG before NG in the root `build.gradle.kts`), NG fails during source patching as well: It loads a class that only exists in the build used by NG, not FG. At least one of these builds is signed, so you end up with classes from jars with different signatures in the same package, which is forbidden and crashes.
  - Note that in both cases the failure occurs *after* the full decompilation, so testing this is rather slow.
- The new `StartEventObserver` is quite hacky, if anyone can suggest a cleaner approach it would be very welcome.

NeoForgeApi/src/main/java/mezz/jei/api/neoforge/NeoForgeTypes.java
```diff
@@ -0,0 +1,35 @@
...
+import net.neoforged.neoforge.fluids.FluidStack;
+
+/**
+ * Built-in {@link IIngredientType} for NeoForge Minecraft.
+ *
+ * @since 9.6.0
+ */
+public final class NeoForgeTypes {
+    /**
+     * @since 9.7.0
```
mezz: have to be careful to update `@since` tags in the api package to match the new version

settings.gradle.kts
```diff
@@ -15,11 +15,21 @@ pluginManagement {
...
 		exclusiveMaven("https://repo.spongepowered.org/repository/maven-public/") {
 			includeGroupByRegex("org\\.spongepowered.*")
 		}
 		gradlePluginPortal()
 	}
+	plugins {
+		id("net.neoforged.gradle.common") version("7.0.80")
+		id("net.neoforged.gradle.userdev") version("7.0.80")
+		id("fabric-loom") version("1.1.+")
+	}
```
mezz: The conflict between NeoForge and Forge may appear because these plugins are being applied at the root project level here. Can these be applied in the NeoForged gradle projects instead? Note there are no FG plugins set up in this file.

mezz (changes requested):
Looking good, I think this is the best approach!

mezz:
I was able to fix these issues, I think. I'll push them to this branch

mezz:
`StartEventObserver` seems good enough for now, it may need patching to NeoForge to improve it.

PR #3413, opened 2024-01-26
[1.16] Only rebuild ingredient filter when client resources reload

embeddedt:
Similar change to #3235, tailored for 1.16. Might also bring this to 1.19.2 at some point if I have time & interest.


mezz:
Thanks!

PR #3431, opened 2024-02-12
Modernize gradle, publish to modrinth, and fix maven publishing

mezz:
Thanks to 
* https://github.com/GTNewHorizons/RetroFuturaGradle
* https://github.com/GregTechCEu/Buildscripts/


settings.gradle.kts
```diff
@@ -0,0 +1,20 @@
...
+                includeGroup("com.gtnewhorizons")
+                includeGroup("com.gtnewhorizons.retrofuturagradle")
+            }
+        }
+        gradlePluginPortal()
+    }
+}
+
+val minecraftVersion: String by settings
+rootProject.name = "jei-${minecraftVersion}"
```
serenibyss: If you use IDEA, in my experience its best to set this like `rootProject.name = rootProject.projectDir.name` (I think that is valid .kts syntax) since sometimes the project structure can get corrupted otherwise

serenibyss (approved):

settings.gradle.kts
mezz: Hmm ok. I use IDEA and this is the name I use in later JEI versions (where it's a gradle multiproject).
What kind of corruption issues did you see?

PR #3469, opened 2024-03-28
Update gradle and related files

ALongStringOfNumbers:
In #3431 the 1.12 version of JEI was updated to use RFG and a newer gradle version. This PR updates gralde again, to 8.5 from 8.4, and uses the command line option to do so, see https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper, which also updates the related gradle files, such as `gradlew` and the wrapper jar.

I noticed that these files were not updated as part of #3431, so I thought that I would just update them here, so that they are no longer the files that were shipped with gradle 4.0.

In addition, this PR switches the gradle distribution from `all` to `bin`, to reduce the download size initially when setting up the project. Let me know if this is not desired.


mezz:
Thanks for contributing!

I trust you but I can't really verify the binary changes, so I think I'll just run the command and submit it

thanks for poking me to update gradle though, I think it makes sense

PR #3582, opened 2024-07-07
bookmark tooltip enhancement

vfyjxf:
Enhanced tooltip for recipe bookmark


Forge/src/main/java/mezz/jei/forge/platform/RenderHelper.java
```diff
@@ -79,4 +81,11 @@ public void renderTooltip(
 			y
 		);
 	}
+
+	@Override
+	public void renderTooltip(Screen screen, GuiGraphics guiGraphics, List<ClientTooltipComponent> components, int x, int y, Font font, ItemStack stack) {
+//		guiGraphics.tooltipStack = stack;
```
vfyjxf: I'm not sure what caused AT to not work on my computer, so I commented it here.

vfyjxf:
preview and ingredients list: 
QQ_1720371269858
preview only: 
QQ_1720371302002
To display the ingredients list you need to press shift.

vfyjxf:
QQ_1720466213688
Also, I added error highlight

mezz:
I have hidden this behind SHIFT by default, and made a configuration option so it can always be shown if you want instead.

mezz:
I set the default config to show only the recipe, not the ingredient summary.
I think that will be the most common and useful option for most people.

mezz:
Thank you for your work on this and contributing a big feature idea to JEI!

PR #3591, opened 2024-07-10
add tag component

vfyjxf:
image


Common/src/main/java/mezz/jei/common/util/TagUtil.java
```diff
@@ -19,9 +19,6 @@ public static <VALUE, STACK> Optional<ResourceLocation> getTagEquivalent(
 		Function<STACK, VALUE> stackToValue,
 		Supplier<Stream<Pair<TagKey<VALUE>, HolderSet.Named<VALUE>>>> tagSupplier
 	) {
-		if (stacks.size() < 2) {
```
vfyjxf, line 22: As strange as it may sound, some of the tags provided by minecraft do have only one item

mezz: Is it useful for players to see tags that have only one item?
I put this filter because I assumed it was not useful, but I can reconsider!
Maybe it can be a config option.

Library/src/main/java/mezz/jei/library/gui/ingredients/TagContentTooltipComponent.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.mojang.blaze3d.vertex.PoseStack;
+import mezz.jei.api.ingredients.IIngredientRenderer;
+import net.minecraft.client.gui.Font;
+import net.minecraft.client.gui.GuiGraphics;
+import net.minecraft.client.gui.screens.inventory.tooltip.ClientTooltipComponent;
+
+import java.util.List;
+
+@SuppressWarnings({"rawtypes", "unchecked"})
```
mezz: we should never need this on a whole class, I can fix it up

Common/src/main/java/mezz/jei/common/util/TagUtil.java
vfyjxf: I think it would be better to display it if the recipe used a tag

mezz: Screenshot 2024-07-12 at 1 29 49 PM
I found some weird tags that are not useful to see, so I will make it configurable

PR #3592, opened 2024-07-11
[1.21] add crafter in "crafting" Recipe Catalyst

AiDSl:
added missing crafter in "crafting" Recipe Catalyst


AiDSl:
closed for added.

PR #3621, opened 2024-07-22
Update Traditional Chinese (zh_tw.json)

dirtTW:
Update and Fixup Traditional Chinese language


mezz:
Thank you for contributing!

PR #3636, opened 2024-07-26
fix fluid rendering in bookmark preview

vfyjxf:
eb26f5295e14f66ec6b47ae7b733c581


vfyjxf:
improve amount display

mezz:
I cleaned this up to merge it, but I'm having a difficult time figuring out how to make the amount render in a way that doesn't expand outside of the fluid.
Either we can render it very small (hard to read), or it overflows...

Screenshot 2024-08-23 at 10 03 26 PM

vfyjxf:
We just need to make his overflow as inconspicuous as possible and keep it readable

mezz:
I think I want to add a "marquee" text for this case, so that the text always fits in a small area but it can scroll over time to show all the information.

mezz:
I think I want to add a "marquee" text for this case, so that the text always fits in a small area but it can scroll over time to show all the information.

vfyjxf:
> marquee

I think marquee text might not be appropriate, at least not for text that goes beyond just a little, where the player might want to determine the exact value at a glance.

vfyjxf:
> I think I want to add a "marquee" text for this case, so that the text always fits in a small area but it can scroll over time to show all the information.

I think marquee text might not be appropriate, at least not for text that goes beyond just a little, where the player might want to determine the exact value at a glance.

PR #3648, opened 2024-08-01
Switch to fastutil maps in key areas of the recipe manager

embeddedt:
This PR improves memory usage of JEI's recipe structures slightly by switching to fastutil maps in a few key places (avoiding the overhead of many `HashMap$Node` objects) and compacting the array lists stored by `uidToRecipes` after recipes are created (to avoid keeping oversized arrays in memory forever).

### DIFF
--- a/Library/src/main/java/mezz/jei/library/load/PluginLoader.java
+++ b/Library/src/main/java/mezz/jei/library/load/PluginLoader.java
@@ -152,6 +152,8 @@ public class PluginLoader {
 		RecipeRegistration recipeRegistration = new RecipeRegistration(jeiHelpers, ingredientManager, ingredientVisibility, vanillaRecipeFactory, recipeManagerInternal);
 		PluginCaller.callOnPlugins("Registering recipes", plugins, p -> p.registerRecipes(recipeRegistration));
 
+		recipeManagerInternal.compact();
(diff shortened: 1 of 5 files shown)


mezz:
Looks good to me! 
Do you have any rough benchmarks for how much this might save in ram, and how much time it takes in a big pack?

embeddedt:
I had actual numbers at the time but forgot to save them (I delete my heap dumps pretty quickly since they take up a lot of disk space). 

> how much this might save in ram

Summing up the total shallow size of `HashMap$Node` objects retained by the `RecipeManager` in the current version of JEI should be a reasonable approximation. The savings were modest in FC but I expect them to be better in a larger pack.

>  how much time it takes in a big pack

Subjectively, I didn't see any increase in JEI loading time when testing this on FC. 

I don't think the fastutil map itself will be significantly slower than a HashMap (it's often equivalent or faster), so the main concern would be the array list trimming, which is a single array allocation + copy of all contents from the previous array. That operation should be quite fast, it doesn't have to deduplicate or anything fancy.

mezz:
Yeah my main concern was the array copy.
I will try this out later but for now it seems like a plausible improvement so I'll merge.
Thanks!

PR #3688, opened 2024-08-14
Update sv_se.json

Regnander:
Added new strings and updated existing ones


mezz:
Thank you for contributing!

PR #3702, opened 2024-08-23
Use Object for UIDs instead of String

mezz:
This will reduce the amount of RAM JEI uses, speed up start times, and improve ingredient comparisons for ingredients with complex subtypes.


mezz:
Tested on a pack with `42,612` ingredients 

before/after sizes 
IngredientFilter 128 MB (no change)
RecipeManager 45 MB (no change)
IngredientManager 4.49 MB -> 2.25 MB

It looks like most of the gains were already realized by the changes here: https://github.com/mezz/JustEnoughItems/commit/7d6649d629b5b621455525f2108d9f1036482234#diff-0cf20bed8b817633503903a8b9d115209d7f5107bd83f863212334ba9b396cd4R27-R35

I think it's still worth going forward with these changes though, because it makes implementing subtype interpreters easier (and more powerful), and using Codecs is much cleaner than relying on UIDs.

PR #3707, opened 2024-08-24
Update Traditional Chinese (zh_tw.json)

dirtTW:
Update Traditional Chinese localization


mezz:
Thank you for contributing!

PR #3712, opened 2024-08-27
Optimize storage of ItemStacks

mezz:
This PR tries to store less data per ItemStack.

The benefit is a bit debatable because ItemStacks need to be created frequently from this data in order to use them for rendering and other purposes.
The memory footprint is lower, but memory churn and cpu usage could be higher.
It'll need some testing to validate whether it's helping or hurting things.


mezz:
There is an additional benefit from this, which is more-immutable ItemStacks.
Based on my tests, there isn't a big performance downside of this PR so I will merge it.

PR #3724, opened 2024-08-31
Update zh_cn.json for 1.21

suoyukii:
Simplified Chinese rework to match the most recent en_us file.


mezz:
Thank you for contributing!

PR #3725, opened 2024-08-31
move the recipe category render call in RecipeLayout

screret:
…to before slots are drawn


mezz:
Thanks for contributing!

Note: this came up in Discord and is a fix for issues with GregTech drawing recipes.
They want to draw arrows and other elements underneath slots, which is currently impossible.

This fix will allow them to do new things without really breaking anything so it seems good to me.

PR #3732, opened 2024-09-02
Use reference set in Node instead of wrapping map

embeddedt:
This saves 5MB in my test pack. Given the IngredientFilter retained about 23MB total before, it's pretty substantial relative to the total size of the data structures.

Performance should be roughly equivalent since `IdentityHashMap` and `ReferenceOpenHashSet` are both using a flat, open addressing table under the hood.


mezz:
Looks good, thanks!

PR #3733, opened 2024-09-02
Optimize AnvilRecipe storage

embeddedt:
This PR aims to heavily compact the storage of `AnvilRecipe` recipes in JEI. The gains are modest with vanilla/smaller modpacks (in my test 1.20.1 pack, it went from 4MB -> 600KB, so it was not that large to begin with) but I expect them to be much more significant with larger packs or with mods that store a lot of NBT on their ingredients, as the number of ItemStack copies made is greatly reduced.

Summary of changes:

1. Enchantment book recipes now store a transforming list that lazily copies the ingredient with the desired enchantment level when the list is accessed. I saw no significant regression to startup performance or UI performance from this change on my fairly weak system. The primary advantage of this change is that we should only be permanently storing one copy of the ingredient regardless of how many enchantment levels or unique enchantments it can have on it.
2. The singleton input ingredient list is now reused across all possible enchantment/enchantment level recipes.
3. The list of enchantment books is now used as-is if the item can be enchanted with every enchantment level (which I suspect is the common case). Fairly simple to do, and should not break anything as it is immutable, anyways.
4. Small cleanup of the repair recipe logic to avoid creating multiple singleton lists with the same element.

To implement 1, `AnvilRecipe` was modified to allow storing the given list in the record directly, rather than copying it to an immutable one, and the defensive copy was moved to the `VanillaRecipeFactory` API. This should not introduce bugs unless mods were using the internal `AnvilRecipe` constructor, and even then, would only cause issues if they were actually mutating their own list. Any mod using the public API still gets a defensive copy made.

### DIFF
--- a/Library/src/main/java/mezz/jei/library/plugins/vanilla/VanillaRecipeFactory.java
+++ b/Library/src/main/java/mezz/jei/library/plugins/vanilla/VanillaRecipeFactory.java
@@ -33,7 +33,7 @@ public class VanillaRecipeFactory implements IVanillaRecipeFactory {
 		ErrorUtil.checkNotEmpty(outputs, "outputs");
 		ErrorUtil.checkNotNull(uid, "uid");
 
-		return new AnvilRecipe(List.of(leftInput), rightInputs, outputs, uid);
+		return new AnvilRecipe(List.of(leftInput), List.copyOf(rightInputs), List.copyOf(outputs), uid);
 	}
 
 	@Override
@@ -42,7 +42,7 @@ public class VanillaRecipeFactory implements IVanillaRecipeFactory {
 		ErrorUtil.checkNotNull(rightInputs, "rightInputs");
 		ErrorUtil.checkNotEmpty(outputs, "outputs");
 
-		return new AnvilRecipe(List.of(leftInput), rightInputs, outputs, null);
+		return new AnvilRecipe(List.of(leftInput), List.copyOf(rightInputs), List.copyOf(outputs), null);
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


mezz:
`List.copyOf` is smart enough to avoid copying immutable lists created with `stream.toList()` and `List.of`. See the implementation, it calls `ImmutableCollections.listCopy`
which looks like 
```java
    static <E> List<E> listCopy(Collection<? extends E> coll) {
        if (coll instanceof List12 || (coll instanceof ListN<?> c && !c.allowNulls)) {
            return (List<E>)coll;
        } else if (coll.isEmpty()) { // implicit nullcheck of coll
            return List.of();
        } else {
            return (List<E>)List.of(coll.toArray());
        }
    }
```

`Lists.transform` does not return a compatible type though.

Given that, I think most of the gains here may come from just a couple places:
* changes to `public List<ItemStack> getEnchantedBooks(ItemStack ingredient)`
* `var ingredientSingletonList = List.of(ingredient);`

(especially in the first case) by sharing the same lists, we save memory.

I am not sure if the rest of the changes are needed for this.

embeddedt:
The change to be able to use the transforming list without copying it is important. That is how we avoid needing to store many copies of the same Item stack with everything identical except the one enchantment. In a modpack with a really expensive tool's NBT/components/etc, you'd end up storing all the tool data 200 times if there are 200 different enchantment/level combos. Now you'd just store it once because all the transforming lists point at the same prototype item and copy it dynamically.

mezz:
Ok that makes sense to me, thanks!

PR #3735, opened 2024-09-03
Reduce initial size of sets for ingredientUidToCategoryMap

embeddedt:
Follow-up to #3648. I noted that many `ObjectOpenHashSet`s were retained here with only 1 element, but had capacity for 16 or more elements, which is rather wasteful. This PR fixes that.


mezz:
Thanks!

PR #3737, opened 2024-09-03
Avoid creating duplicate ResourceLocation objects for anvil enchantment recipes

embeddedt:
This saves a couple more megabytes in ATM9NF.

Note: I found it strange that the UID does not include the specific enchantment and level (which would make it truly unique between the recipes). However, I opted to exploit that for further memory savings, rather than fix it to be unique, because I assume it was done this way for a reason, or has been like this for a while, and therefore no one is relying on it being unique among enchantment/level combos (only among unique input ingredients).

If the intent is for the UID to actually be unique for that exact recipe, this optimization can't be made. However, do note that on 1.20.1, the input ingredient's NBT is often stringified and used as part of the name - so this will again cause exponential growth if items with large NBT get enchanted.

### DIFF
--- a/Library/src/main/java/mezz/jei/library/plugins/vanilla/anvil/AnvilRecipeMaker.java
+++ b/Library/src/main/java/mezz/jei/library/plugins/vanilla/anvil/AnvilRecipeMaker.java
@@ -115,16 +115,16 @@ public final class AnvilRecipeMaker {
 		ItemStack ingredient
 	) {
 		var ingredientSingletonList = List.of(ingredient);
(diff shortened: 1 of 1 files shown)


mezz:
Ah unfortunately instead of an optimization opportunity, you have uncovered a bug.
If I try to bookmark the fortune recipe (top) it also treats the mending recipe (bottom) as being bookmarked because the id matches.

Screenshot 2024-09-04 at 10 40 48 AM

Screenshot 2024-09-04 at 10 41 44 AM

mezz:
I have removed the ability to bookmark these specific recipes in the latest version here:
* https://github.com/mezz/JustEnoughItems/commit/5e9ec3620726ab88d5aee3b5f4dad9f043901e6b

PR #3739, opened 2024-09-04
Switch to new ModDevGradle plugin

Technici4n:
Note that MDG supports parchment, if you want to use it in all projects (especially in common). See https://github.com/neoforged/ModDevGradle?tab=readme-ov-file#better-minecraft-parameter-names--javadoc-parchment.

### DIFF
--- a/NeoForgeApi/build.gradle.kts
+++ b/NeoForgeApi/build.gradle.kts
@@ -54,6 +54,7 @@ dependencies {
 
 neoForge {
 	version = neoforgeVersion
+	// We don't need the AT, but this allows MDG to share the recompiled Minecraft artifacts with the NeoForge project.
 	accessTransformers.from(file("../NeoForge/src/main/resources/META-INF/accesstransformer.cfg"))
 }
 
--- a/NeoForgeApi/build.gradle.kts
+++ b/NeoForgeApi/build.gradle.kts
@@ -55,7 +55,7 @@ dependencies {
 neoForge {
 	version = neoforgeVersion
 	// We don't need the AT, but this allows MDG to share the recompiled Minecraft artifacts with the NeoForge project.
-	accessTransformers.from(file("../NeoForge/src/main/resources/META-INF/accesstransformer.cfg"))
+	setAccessTransformers("../NeoForge/src/main/resources/META-INF/accesstransformer.cfg")
 }
 
 val sourcesJar = tasks.named<Jar>("sourcesJar")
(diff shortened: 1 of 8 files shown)


NeoForgeApi/build.gradle.kts
```diff
@@ -47,22 +47,14 @@ java {
...
 
-minecraft {
-	accessTransformers {
-		// All minecraft configurations in the multi-project must be identical, including ATs,
-		// because of a ForgeGradle bug https://github.com/MinecraftForge/ForgeGradle/issues/844
-		file("../Forge/src/main/resources/META-INF/accesstransformer.cfg")
-	}
+neoForge {
+	version = neoforgeVersion
+	accessTransformers.from(file("../NeoForge/src/main/resources/META-INF/accesstransformer.cfg"))
```
mezz: NeoForgeApi doesn't actually need ATs, this was just included as a workaround for https://github.com/MinecraftForge/ForgeGradle/issues/844

Is this still needed with MDG?

build.gradle.kts
```diff
@@ -4,9 +4,7 @@ plugins {
     // https://maven.fabricmc.net/fabric-loom/fabric-loom.gradle.plugin/
     id("fabric-loom") version("1.7.3") apply(false)
     // https://projects.neoforged.net/neoforged/neogradle
```
mezz:
```suggestion
    // https://projects.neoforged.net/neoforged/moddevgradle
```
these URLs are intended to point to the place to find the latest version numbers (like `2.0.26-beta` here)

mezz (changes requested):
Thanks for trying it out. I'm surprised it worked easily.

Were you able to test that the output jars are the same?

Having parchment would be nice. I had it set up before but ran into issues.

NeoForgeApi/build.gradle.kts
Technici4n: It's not needed, no. I left it in to save one recompilation, but removing it is a bit cleaner. (Once it's in the cache it doesn't matter for subsequent runs.)

Technici4n:
> Were you able to test that the output jars are the same?

Yes, they are the same when published to maven local at least!

NeoForgeApi/build.gradle.kts
mezz: We can leave it in, that sounds good. Can you add a comment above this line to explain why it's here?

mezz (approved):

PR #3780, opened 2024-09-27
U.S. English translation revamp and cleanup of other lang files

Greg-21:
The changes I've made include various improvements to the source U.S. English translation, enhancements to the comments in the configuration files to make them at least a bit more clear and helpful (I hope), and last but not least, an overall cleanup of the rest of the translations to align them better with the current U.S. English translation and make it easier for future translators to pick up where their predecessors left off and continue from there.

I realize that what seems like a positive change to me, may actually be the opposite, so I'm open to all kinds of criticism and comments. After all, this is your mod and your repository (not mine), and without it, I could do nothing more than flip through pages for 40 days and 40 nights in the Risugami's Recipe Book like in the good old days :)

This pull request consists of three main parts (one commit for each of these points):
1. U.S. English translation revamp:
   - Some of the sentences used in tooltips and various messages have been rephrased in the hope of making them clearer and easier to read, as well as easier to translate into other languages reliably.
   - The "compostable" recipe category, previously described by an adjective (which was an exception), has now been replaced by the gerund "composting" to better match the other categories like "smelting" or "brewing".
   - Another thing worth noting is a capitalization of key bind descriptions in the key bind configuration screen, as well as an addition of periods at the ends of sentences in tooltips for parity with vanilla Minecraft.
   - In addition, the style of sentences informing the user to press something for an action to happen has been standardized across all occurrences.
   - Last but not least, the following two translation keys: `jei.tooltip.error.recipe.transfer.unknown` and `jei.tooltip.error.recipe.transfer.disabled` have been removed, as they haven't been in use for 8 (!) and 3 years respectively. The code associated with the former has been removed in commit 9097e16723308d79acf9fef5cdb54604231502a5, and the code for the latter in commit f8e5522d851126d14c716254a4a6e7a524eb47ce (only 20 minutes after the feature was introduced, just a little fun fact).
2. Configuration file comment revamp:
   - This includes unifying the descriptions of various configuration options in the config files with those in the U.S. English translation file where possible. Since some of them are based on the rephrased sentences from point 1, it will hopefully make them a bit more clear and helpful.
   - It also adds periods at the ends of comment sentences, which is something you've already done a few times recently when adding new configuration options, and I've just picked it up and applied it consistently across all the sentences in the configuration files. Thanks to this change, the automatically generated in-game tooltips by MrCrayfish's Configured mod for the various options in the configuration screen will also end with periods, thus matching the descriptions of the vanilla Minecraft options added by Mojang itself.
3. Translation file cleanup:
   - This includes rearranging all lines so that they are in the same order as in the source U.S. English translation file, removing translation keys that are no longer in use (see point 1 above for the list of what they are), and unifying line endings to always use LF only. Most of this, if not all, was done automatically by a script I created, so the risk of error is minimal.
   - Another thing worth noting is the removal of lines that were left untranslated, in English, often containing long-outdated option descriptions, tooltips, button labels, etc. (this is basically doing the opposite of what you used to do in the past, for example in commit 325190132a44ba69e2cb102420717ca3cc3f2ff0). This should make it easier for future translators to pick up where their predecessors left off and continue from there.
   - It also renames the file containing the Norwegian Bokmål translation from `nb_no.json` to `no_no.json‌`, since the latter is the correct locale code used in Java Edition for this language. It has been the case for a while now, but apparently no one has noticed, or the absence of this translation hasn't been significant enough for anyone to contribute a fix.
   - Complete removal of the Australian English translation, as it has been unmaintained and hasn't received any updates for almost 8 years (!). For people using this language, it will be much better to use the U.S. English translation as a fallback instead.

The list above is not exhaustive and doesn't cover all the changes, only the most significant ones (and even then, it's long enough that a TL;DR would be handy, so I'm not making it any longer).


Common/src/main/resources/assets/jei/lang/en_us.json
```diff
@@ -126,38 +124,38 @@
...
   "config.jei.advanced.maxColumns": "Max Overlay Width",
   "config.jei.advanced.maxColumns.comment": "The maximum width of the ingredient and bookmark list overlays.",
   "config.jei.advanced.maxRecipeGuiHeight": "Max Recipe GUI Height",
   "config.jei.advanced.maxRecipeGuiHeight.comment": "The maximum height of the recipe GUI.",
   "config.jei.advanced.giveMode": "Give Mode",
-  "config.jei.advanced.giveMode.comment": "Choose if JEI should give ingredients direct to the inventory (inventory) or pick them up with the mouse (mouse_pickup).",
+  "config.jei.advanced.giveMode.comment": "Choose if JEI should give ingredients directly to the inventory (inventory) or pick them up with the mouse (mouse_pickup).",
   "config.jei.advanced.addBookmarksToFront": "Prepend New Bookmarks",
-  "config.jei.advanced.addBookmarksToFront.comment": "When true, adds new bookmarks to the front of the bookmark list. When false, adds new bookmarks to the end of the bookmark list",
+  "config.jei.advanced.addBookmarksToFront.comment": "Add new bookmarks to the front of the bookmark list instead of the end.",
```
Greg-21, line 153: This entire section also qualifies for deletion, as these text strings don't seem to have been used since the release of JEI for Minecraft 1.13. However, since it's quite a large chunk, I preferred not to touch it before consulting with you first. What's your opinion on this?

mezz: I forgot to implement support for these in the new config system.
I need to add support for localization in client configs, so let's leave these here.

mezz:
Thank you for diligently going through to fix these up, and documenting your changes in the PR description here.
Rest assured your effort is appreciated!

I looked through and the changes all look very positive (or at least neutral), so I have no problem merging these in. Thanks again!

Greg-21:
Would you be interested in having me port this fairly large set of changes to the older Minecraft versions you still support?

It wouldn’t be much of an issue on my end, as most of the porting process can be automated, and I already have the required scripts for this task.

mezz:
Thank you, that would be appreciated.

mezz:
@Greg-21 I have re-implemented the localization for config files in this commit: https://github.com/mezz/JustEnoughItems/commit/719c00ecc104fa53d46db0fc7dc6f02f246385cc

I haven't removed all the old unused values yet, because I may need them when I implement a config GUI soon.
I just wanted to give you a heads up in case you were planning on doing anything with them.

Greg-21:
As far as I'm concerned, feel free to remove whatever you want. If I need anything in the future, I can always branch off from *any* commit and start from there.

EDIT: Now that you mention it, I think I'll also remove all the remnants of the old in-game configuration interface from the older Minecraft versions if you plan to do the same in the current one.

PR #3801, opened 2024-10-07
add history overlay

vfyjxf:
image

aabae388102fbfcde269106a61db4d21

add a view history feature to jei.


castilio20:
1.a crafting history would be nice to
2.config option to let the player chose where the history is going to be placed aka left above the inventory or on the right

Yamijiro:
Yes please. A recipie viewing history would be insane so users don't have to bookmark every step of a bigger craft.

vfyjxf:
> Yes please. A recipie viewing history would be insane so users don't have to bookmark every step of a bigger craft.

At the moment, this feature is really more of a “lookup history” than a history of the recipes I've seen, and I'm looking for a clean interface to display the history of the recipes.

PR #3808, opened 2024-10-12
add control to force potentially existing bookmark to top of list

Goldenlion5648:
I am the type of player that bookmarks items, but doesn't remove them once done with that item (or the item is used through out the pack). This becomes annoying when I want to put an item near the top of the list (the setting I use), but it is already in the bookmark list.

This PR adds the "shift + a" keybind, which removes an existing bookmark, and adds it back. If the item was not already bookmarked, it is added as if the normal "a" keybind was pressed

### DIFF
--- a/Common/src/main/java/mezz/jei/common/input/IInternalKeyMappings.java
+++ b/Common/src/main/java/mezz/jei/common/input/IInternalKeyMappings.java
@@ -23,6 +23,7 @@ public interface IInternalKeyMappings extends IJeiKeyMappings {
 	IJeiKeyMapping getCloseRecipeGui();
 
 	IJeiKeyMapping getBookmark();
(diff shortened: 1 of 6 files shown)


vfyjxf:
Why can't you try dragging the bookmark you want to move to the top, jei supports that.

Goldenlion5648:
> Why can't you try dragging the bookmark you want to move to the top, jei supports that.

That still requires finding the item in the bookmarks list, which can be annoying if there are multiple pages of bookmarks

Goldenlion5648:
Oh wait this currently doesn't work when "shift a" is pressed on an item in the bookmarks (as in hovering an item on the left side of the inventory), it just removes it. New PR incoming

Goldenlion5648:
Any update on if this can be merged?

PR #3799, opened 2024-10-05
Follow-up work to revamp the U.S. English translation and clean up other lang files

Greg-21:
As the title suggests, this is a follow-up work to revamp the U.S. English translation and clean up other language files. It primarily focuses on fixing a stupid mistake I made in #3780, related to the following text from the linked PR:
> This includes unifying the descriptions of various configuration options in the config files with those in the U.S. English translation file where possible. (...)

For some reason, instead of replacing the older phrase "resource IDs" with the newer and more accurate "resource locations", I did it the other way around. So now I'm fixing that and doing what I should have done from the very beginning, along with other minor cleanups here and there.

I apologize for making such stupid mistakes.

PR #3809, opened 2024-10-12
[1.20.6] U.S. English translation revamp, backport, and cleanup of other lang files

Greg-21:
This is basically a backport of PR #3780 and PR #3799, as well as the recently accepted and merged Japanese and Polish translations, which unfortunately didn't make it into that cleanup. For a detailed list of changes, please refer to the linked pull requests.

Of course, due to the differences (sometimes significant) between JEI for 1.21.1 and 1.20.6, not everything could be backported, and even if it could, it didn't always make sense. For example, almost every time I switched to an older version, there were some differences in the configuration options offered by JEI. So, for obvious reasons, there was no point in backporting descriptions for options that simply don't exist in that older version.

In addition, there's also something new here — the removal of all remnants of the old in-game configuration interface from translation files. This interface hasn't been in use for years, the configuration option descriptions are outdated, and some of them don't even correspond to any available options in config files anymore, while others are just missing. If the in-game configuration interface were to be reintroduced in JEI for this Minecraft version, it would probably make more sense to backport everything from 1.21.1 instead.


Greg-21:
I apologize that it took me longer than I initially anticipated. But in the meantime, I decided to do a clean installation of the latest Windows 11 24H2 update that was released earlier this month and upgrade from 23H2, as Microsoft would have done it for me sooner or later anyway. Now at least I had full control over what was done and how.

mezz:
Thank you!

PR #3834, opened 2024-10-29
add history overlay

vfyjxf:
Same as #3801, except that pr was turned off because I incorrectly updated the downstream


ghost (approved):
Looks absolutly fine

juzi5201314:
It would be better if there is an option to place it on the right side instead of the left side

vfyjxf:
> It would be better if there is an option to place it on the right side instead of the left side

It should be configurable in the future, but in the beginning mezz told me it would be better on the left side, since the browsing history obviously has little to do with the list of browsed material, it's more like a bookmark of a temporary nature.

shadessass:
release?

mezz:
Thanks for contributing!

I did some polishing to complete this feature:
* created icons
* fixed a bug where lookups weren't saved if they came from vanilla guis using R/U hotkeys
* removed the line animation and made the line darker so it is less distracting
* made sure the dashed lines have complete dashes and don't get cut off partway
* added description string for the button
* ensure the layout updates immediately when the configs change, to avoid overlaps and visual glitches
* save the history to config file so that it persists between play sessions
* localized configs

vfyjxf:
Great changes, but I'm a little curious as to why you don't allow lookup history on both sides.

mezz:
I think when implementing things like this, it is important only add options that give a large benefit and are easy to understand.

Having the lookup history on both sides is not very consistent with other GUIs in Minecraft or in JEI, where one thing is always in one place. 
If we allow the option for both, the GUI is no longer "normal" and may need some additional explanation for people to understand.
I do not see a very large benefit for having it on both sides, so I removed the ability in order to reduce complexity.

PR #3889, opened 2025-01-12
1.20.1 JEI async recipe loading and reloading

Mysticpasta1:
Pulled and updated the async-load Branch to 1.20.1 forge and fabric


mezz:
Sorry I appreciate your effort but I cannot accept this PR. 
There are breaking changes to the API and in my testing, there were many subtle issues that could come up from mods that are not expecting concurrency when loading. I think the number of issues from mods that have problems would be too high.
I may revisit this idea for later versions but for now I do not think there is a safe way forward.

vfyjxf:
Would it be feasible to put it in 1.22 as a breaking change?

PR #3890, opened 2025-01-13
Fix translation in zh_cn.json

YocyCraft:
The translation of `"jei.tooltip.recipe.tag"` (en_us: `Accepts tag: %s`) in zh_cn  is `"接受下列任何物品：%s"`, which means "Accept any of the following **items**: %s" literally. However, in the game this translation key can be used for any recipes with tagged ingredients, including items, fluids or any custom tagged ingredients. This could sometimes confuse players when they see a non-item ingredient with this tooltip. The new translation means "Accept **ingredients** in the tag: %s" which could make better understanding.

image


mezz:
Thank you!

PR #3911, opened 2025-02-05
Don't draw tooltips in exclusion areas

ALongStringOfNumbers:
Prevents JEI tooltips from being drawn in zones marked as exclusion zones. Prevents issues with some moveable GUI elements and JEI tooltips.


CursedGhoul (approved):

mezz:
Thank you for contributing!

Normally tooltips draw over all other GUI elements, even in exclusion areas. I do not see an issue with the current behavior, can you please explain and demonstrate the issue and the resolution with screenshots?

PR #3985, opened 2025-05-10
made CraftingGridHelper.getCraftingIndex better readable 

Nischhelm:
technically allows arbitrary height + width of the crafting grid but ofc that helper is for vanilla crafting tables. only needs to know the input recipe width and the crafting grid width, heights get auto aligned.


Nischhelm:
nvm i didnt realise it doesnt align top left, will check if i can fix.

Nischhelm:
problem is that the JEI alignment doesnt follow a specific overarching rule. 3x2 aligns bottom, 2x2 aligns top left. Vanilla doesnt follow a specific rule either though, with lots of recipes aligning top, some center. maybe better to just leave this alone, sry for the bother

Issue #3981, opened 2025-05-01
[Bug]: Vanilla search bar has priority over JEI search bar


Krampus69:
### Steps to Reproduce the Bug

Hello ! The JEI search bar in creative tab cannot be used if the vanilla search bar is visible, you have to manually change the page to something else everytime to be able to type something, which after months of being like this has got so annoying that I had to make a ticket for it

Here is a video to explain the issue : https://streamable.com/u1lmff

Minecraft 1.20.1
Forge 47.4.0
jei-1.20.1-forge-15.20.0.106

### Expected Behavior

Clicking on the search bar should allow me to type something in it, even if the vanilla search bar is open

### Actual Behavior

It just doesn't, you have to change the creative tab page to type something

### Mod Pack URL (Optional)

https://www.curseforge.com/minecraft/modpacks/minecraft-legendary-edition

### Mod Pack Version (Optional)

2.12

### latest.log

https://mclo.gs/C5rDTYf

PR #4001, opened 2025-05-30
Fix vanilla/JEI search bar focus conflict

lexipagliaro:
Aims to fix the issue raised in #3981, making it possible to select and type into the JEI search bar even when the vanilla search tab is open. Clicking away from the JEI search bar returns focus to vanilla.

image

This is my first attempt at contributing to a project and I tried my best to understand the process and code--I apologize if I've gone about things incorrectly or implemented inefficiently! Let me know if there is anything I can do. I appreciate your work :)


mezz:
Thank you for contributing!
There is nothing wrong with your approach, but based on what you write I wanted to handle this more generally so I implemented it here: https://github.com/mezz/JustEnoughItems/commit/10c3bbb26339247d969b0a37f7ad9501d56d1bb6
Thank you for the idea.

PR #4023, opened 2025-06-20
Update uk_ua.json

StarmanMine142:


Common/src/main/resources/assets/jei/lang/uk_ua.json
```diff
@@ -178,4 +331,6 @@
   "description.jei.debug.formatting.3": "%s nested",
   "jei.alias.panda.spawn.egg": "endangered",
   "jei.alias.villager.spawn.egg": "HMMM"
+  
+  "modmenu.descriptionTranslation.jei": "JEI — це мод для перегляду предметів і рецептів майстрування, створений з нуля для стабільности та продуктивности",
```
Greg-21, line 332: Shouldn't these types of translation keys that are only applicable to specific platforms be placed in their designated directories (in this case, Fabric and Quilt with the Fabric compatibility layer, since Prospector's Mod Menu has always been a Fabric-exclusive mod) rather than in the shared "common" directory to avoid unnecessary clutter and confusion for future translators who have never used Mod Menu and/or Fabric?

This isn't necessarily something you need to address. I'm mainly flagging it for @mezz, as it might be something to consider at the repo level.

mezz: I agree it would be technically more correct to split the files for each mod loader, but for practical purposes I think it is simpler to include the keys together. If there are any keys that you are unsure of, please feel free to ask me about it.

mezz:
Thank you!

Common/src/main/resources/assets/jei/lang/uk_ua.json
Greg-21: On second thought, I agree that it probably doesn't make much sense to split translations into separate parts for each mod loader while only Mod Menu and Fancy Mod Loader support this kind of mod description translation, and others like LexForge, Quilt, and the rest haven't come up with their own equivalents yet.

> If there are any keys that you are unsure of, please feel free to ask me about it.

@mezz, if I ever run into any issues with that, I definitely won't hesitate to ask for help. Fortunately, JEI is pretty straightforward in this regard, so I haven't really had any trouble with it so far (unlike (Neo)Forge, for example, where most of the messages are some kind of obscure mystery).

I was mainly bringing this up with less experienced Minecraft mod translators in mind, who aren't yet familiar with all the nuances of translating Minecraft mods. And I'm not just speaking hypothetically here. In case you haven't noticed, a few translations that were merged fairly recently are still stuck in a limbo called the 1.21.x branch, and no one seems to have noticed that those new or updated JEI translations aren't actually available in the game, not even their original authors (!). Otherwise, they would probably be complaining about it in the GitHub Issues section or submitting new PRs targeting a different branch...

PR #4038, opened 2025-07-10
Add default grindstone category

Sleepy-Horse:
Add #1776 
The category looks like this:

PR #4050, opened 2025-07-18
Add grindstone category [1.21.1]

Sleepy-Horse:
Port of #4038 to 1.21.1
Confirmed working with vanilla and modded recipes.
Screenshot:

I couldn't test it on fabric or forge because I couldn’t build them. If the problem is on my side, please let me know and I’ll try to fix it.


mezz:
Thank you!

PR #4094, opened 2025-09-24
optimize history behavior

vfyjxf:
- (~~add recipe bookmark to lookup history~~)add recipe bookmark focus value to lookup history
- history button is able to click when bookmrk overlay isn't displayed
~~set default history side to right(because bookmark overlay must be not empty to display history overlay)~~

### DIFF
--- a/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
+++ b/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
@@ -122,7 +122,7 @@ public final class ClientConfig implements IClientConfig {
 		);
 		lookupHistoryDisplaySide = lookupHistory.addEnum(
 			"displaySide",
-			HistoryDisplaySide.LEFT
+			HistoryDisplaySide.RIGHT
 		);
 
 		IConfigCategoryBuilder advanced = schema.addCategory("advanced");
--- a/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
+++ b/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
@@ -122,7 +122,7 @@ public final class ClientConfig implements IClientConfig {
 		);
 		lookupHistoryDisplaySide = lookupHistory.addEnum(
 			"displaySide",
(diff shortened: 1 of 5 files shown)


mezz:
Thank you for contributing!

> add recipe bookmark to lookup history

I left this out because I do not think it's very useful in practice. I tested it out in a pack to see how it feels.
What I found is: 
* If I look up the recipe for something in a normal mod pack, I get many results so it is not usable in that case (and this behavior is not intuitive to players).
* but even if there is only one result, the lookup will be added to the history already, so why is the result also needed? it seems kind of like a duplicate

> history button is able to click when bookmrk overlay isn't displayed

Thanks, this is a bug that need to be fixed.

> set default history side to right(because bookmark overlay must be not empty to display history overlay)

This is not the right fix for the problem, it should be possible to show the history overlay even when the bookmark overlay is empty.

vfyjxf:
> * add recipe bookmark to lookup history

I just noticed that lookup will not be add to history when open from recipe bookmark, i think it was a bit weird. So i add this and make it more powerful.

> * set default history side to right(because bookmark overlay must be not empty to display history overlay)

After reconsidering, I think you're right, because the default placement on the right side takes up valuable space in ingredient overly, but generally speaking bookmark overly has more free space

I will fix bookmark overly later.

mezz:
> I just noticed that lookup will not be add to history when open from recipe bookmark

Ah I see, history from opening a recipe bookmark makes sense to me. 
I think the feature should be limited to do that, so that it is simple to understand.

> After reconsidering, I think you're right, because the default placement on the right side takes up valuable space in ingredient overly, but generally speaking bookmark overly has more free space

I think the problem is not really a debate between left vs right, it's just that even if you set the default to the right side, someone can configure it to the left and will still encounter the bug.

vfyjxf:
ok, i finished those changes, please check first message

mezz:
Looks good!
I made one small improvement, which is to create a recipe bookmark in the history when you lookup a recipe.

PR #4095, opened 2025-09-25
add ingredients summary to recipe output slot's tooltip

vfyjxf:
preview:

### DIFF
--- a/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
+++ b/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
@@ -40,6 +40,9 @@ public final class ClientConfig implements IClientConfig {
 	private final ConfigValue<Integer> maxLookupHistoryIngredients;
 	private final ConfigValue<HistoryDisplaySide> lookupHistoryDisplaySide;
 
+	// recipes gui
+	private final ConfigValue<Boolean> ingredientsSummaryEnabled;
+
 	// advanced
 	private final Supplier<Boolean> lowMemorySlowSearchEnabled;
 	private final Supplier<Boolean> catchRenderErrorsEnabled;
@@ -124,6 +127,9 @@ public final class ClientConfig implements IClientConfig {
 			"displaySide",
 			HistoryDisplaySide.LEFT
 		);
+
+		IConfigCategoryBuilder recipesGui = schema.addCategory("recipesGui");
+		ingredientsSummaryEnabled = recipesGui.addBoolean("ingredientsSummaryEnabled", true);
 
 		IConfigCategoryBuilder advanced = schema.addCategory("advanced");
 		catchRenderErrorsEnabled = advanced.addBoolean("catchRenderErrorsEnabled", !isDev);
@@ -257,6 +263,11 @@ public final class ClientConfig implements IClientConfig {
 	@Override
 	public void addLookupHistoryDisplaySideListener(IConfigListener<HistoryDisplaySide> listener) {
 		lookupHistoryDisplaySide.addListener(listener);
+	}
+
+	@Override
+	public boolean isIngredientsSummaryEnabled() {
+		return ingredientsSummaryEnabled.get();
 	}
 
 	@Override
--- a/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
+++ b/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
@@ -97,6 +97,7 @@ public final class ClientConfig implements IClientConfig {
 		showCreativeTabNamesEnabled = tooltips.addBoolean("showCreativeTabNamesEnabled", false);
 		tagContentTooltipEnabled = tooltips.addBoolean("tagContentTooltipEnabled", true);
 		hideSingleTagContentTooltipEnabled = tooltips.addBoolean("hideSingleTagContentTooltipEnabled", true);
+		ingredientsSummaryEnabled = tooltips.addBoolean("enableRecipesGuiIngredientsSummary", false);
 
 		IConfigCategoryBuilder performance = schema.addCategory("performance");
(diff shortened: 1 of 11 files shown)


mezz:
I fixed the localization and put it in the "tooltips" section of the configs.

I made this disabled by default because it is a power-user feature. There is redundant information displayed, so I want to make it opt-in for people who will understand.

vfyjxf:
> I fixed the localization and put it in the "tooltips" section of the configs.
> 
> I made this disabled by default because it is a power-user feature. There is redundant information displayed, so I want to make it opt-in for people who will understand.

i think we can add a button to quick enable/disable this feature, jei doesn't have a config panel,player can't find those feature easily

Greg-21:
The built-in configuration screens that we all know so well from the older versions of Forge are back in NeoForge 21.0.110+ ([the announcement](https://neoforged.net/news/21.0release/#config-gui-returns-added-in-neoforge-210110-beta)), so all that's needed now is to hook them up on the JEI's side and drop the (broken) integration with the third-party Configured mod.

<sub>Then I, as a translator and the current maintainer of the Polish localization for JEI, will be able to start re-translating all the configuration options, just like in the good old days, which I haven't done yet simply because I had no way to test my translation in-game and ensure it all fits together properly.</sub>

vfyjxf:
> The built-in configuration screens that we all know so well from the older versions of Forge are back in NeoForge 21.0.110+ ([the announcement](https://neoforged.net/news/21.0release/#config-gui-returns-added-in-neoforge-210110-beta)), so all that's needed now is to hook them up on the JEI's side and drop the (broken) integration with the third-party Configured mod.
> 
No, jei use its own config system,and neoforge config screen requie neoforge config system. so we can't use neo config screen,unless we provide different config impl for diffferent platform.

I'm working for a lightweight ingame config context menu

Greg-21:
Instead of reinventing the wheel by creating yet another custom configuration GUI and asking users to learn how to use it, wouldn't it be more practical to rely on existing solutions, such as the aforementioned NeoForge Configuration API on NeoForge and the [Forge Config API Port](https://github.com/Fuzss/forgeconfigapiport) on Fabric/Quilt (Jar-in-Jar'd or as an optional dependency)?

Modern JEI already makes use of the TOML-based NeoForge Configuration API for server-side configs, and the same approach could also be extended to client and common configs (just like in the good old days, when we had `.cfg` files provided by Forge basically everywhere, along with out-of-the-box support for in-game config GUIs). This would give text editor users (e.g. modpack creators) a unified, consistent, and well-established way to (pre)configure JEI, while also introducing simple screens for GUI users, like the one shown below.

Server-side-JEI-config-utilizing-NeoForges-configuration-GUI

vfyjxf:
Jei may have created its own config library to minimize dependencies, I don't think it makes sense to provide full functionality on one platform, and using jij may not be a good idea.
What about your opinion,@mezz

mezz:
I need the config code to work on every modloader, not just NeoForge.
I have looked into how to make it work with NeoForge's config GUI, but unfortunately their code isn't really extensible.
I'm seeing if I can refactor it to make it extensible, or make my own config GUI, but it's a lot of work.
The alternative is to implement configs differently for every mod loader, which is also a lot of work and even more maintenance.

vfyjxf:
I'm adding a simple in-game configuration context menu, it's limited to configuring jei client and jei debug, it shouldn't be very complicated and since it's just ui, it doesn't require any platform specific code

Greg-21:
We kinda got sidetracked by all those discussions about configuration stuff, and because of that, there's one important question I've completely forgotten to ask, even though I really should have. Do you plan to port this new feature to the latest version of Minecraft as well, or will it remain exclusive to 1.21.1?

vfyjxf:
> We kinda got sidetracked by all those discussions about configuration stuff, and because of that, there's one important question I've completely forgotten to ask, even though I really should have. Do you plan to port this new feature to the latest version of Minecraft as well, or will it remain exclusive to 1.21.1?

I usually develop based on 1.21.1, but I'm not familiar with newer versions, I'm not sure exactly which version to port it to, I don't want to create more than a dozen pr....

mezz:
The right approach is to target the "latest" version and then backport to whichever version you like.
In this case, to make sure this doesn't get lost on future versions, the best thing to do is to port to 1.21.10.

I don't think many people are using the versions in between 1.21.1 and 1.21.10 anyway.

Greg-21:
If I may put my two cents in, it might be a good idea to revive the `1.21.x` Git branch as a safe target for all upcoming changes and pull requests. That would save everyone the hassle of constantly switching the PR target branch to yet another version number just to keep up with Mojang's new, faster release cycle.

This is pretty much the approach the (Neo)Forge team and a bunch of other modders have settled on, and it seems to work great in practice.

And if you wanted to take it a step further and get extra fancy, you could also check out henkelmax's branch naming convention for [his GitHub repositories](https://github.com/henkelmax?tab=repositories) related to his Minecraft mods. He has a super clean system that makes it immediately obvious which Minecraft versions are currently supported and which ones are not anymore. That kind of clarity is extremely helpful, as it saves a lot of time and prevents wasted effort on contributions that would otherwise end up being rejected because, well... that particular version just isn’t supported anymore.

mezz:
Thank you for the suggestions, @Greg-21 , I think they're mostly good.
The `1.21.x` branch style was more useful when breaking changes from Mojang happened on `1.x`, but now that breaking changes are happening all the time, it's a bit broader than I would like.

I could keep a `latest` branch, but then managing the build version in CI becomes complex. Each CI build pipeline (one for each branch) has a counter that is used for the last "number" of the version. If I change the names of the branches, that counter gets reset, potentially leading to multiple releases with the same version, and it's hard to keep track of manually.

To adopt a simpler branch naming convention, I will need to fix my build system counter approach first.

PR #4106, opened 2025-10-06
Config screen

vfyjxf:
current preview:
image

This PR is still very early, the code is messy, and the context menu doesn't have any functionality, and there is no support for list type configuration, and there is no separate texture created for the sake of testing other UI textures. But other parts are already working and I would like to get some feedback on whether a CategoryTab-based ui like RecipesGui would be better, but I'm not very good at drawing because the jei client config has too many Categories, so I used the current design.

The scrollbar also needs to add mouse drag support and may need to support smoother scrolling

There is also the compatibility of the config UI with other systems, for example we may need to add a lot of config listeners to respond to configuration changes, or add a discard/apply button to let the player decide


vfyjxf:
Since I don't have much time to draw various textures and work with color scheme themes, it looks ugly :P

mezz:
Screenshot 2026-07-14 at 1 34 39

Screenshot 2026-07-14 at 1 46 07

making some progress on polishing!

mezz:
Still working hard on this but I am making a mess of the commits, I will submit them after pulling some big stuff out and putting them directly on the 1.21.1 branch first

vfyjxf:
replaced by MezzConfigGui

PR #4116, opened 2025-10-19
Collate various potion reagents into single recipe (#4035)

### LINKED ISSUE #4035: [Bug]: JEI Assumes All Potion Types Have One Recipe

1. Clone Natural Philosophy (found [here](https://github.com/voidsong-dragonfly/naturalphilosophy))
2. Run `./gradlew runClient`
3. Check JEI for recipes for the `Water Breathing` potion, only `Pufferfish` will be shown as an ingredient.

### Expected Behavior

The recipe using `Red Algae` from Natural Philosophy to craft `Water Breathing` potions should also be shown in JEI, alongside the `Pufferfish` recipe.

### Actual Behavior

Only one (the first registered) ingredient for a potion type will be shown in JEI, instead of all reagents for a specific potion type. 

### Mod Pack URL (Optional)

_No response_

_No response_

### Extra Notes (Optional)

Bug is present in both `19.21.0.247` and `19.21.2.313` versions of JEI on NeoForge.

### latest.log

https://gist.github.com/voidsong-dragonfly/a653d59384638563374fc91cf514e9b4

noobanidus:
This PR fixes #4035 

JeiBrewingRecipe instances created by BrewingRecipeMakerCommon are given a uid that combines purely the input potion type and the output potion type.

This uid is used in equality checks, which means that additional recipes for a potion with an existing recipe are completely ignored.

Instead of modifying the uid to include the reagent, this instead collates existing reagents and, if the reagent of the "new" recipe isn't already contained, replaces the existing recipe with a new one that includes all previous reagents.

This seemed a better option and more inline with how JEI already handles inputs that produce the same output than creating separate recipes.

There could potentially be an issue with custom implementations of IJeiBrewingRecipe, specifically those that directly subclass JeiBrewingRecipe, and do "complicated things" with ingredients, as those would be replaced with a standard JeiBrewingRecipe. I'm not sure if it's an edge case that can occur to require direct checking. All other implementations would fail JeiBrewingRecipe's equality check.

Finally, my only remaining concern with this solution is `BrewingRecipeUtil::addRecipe` which is manually called by `JeiBrewingRecipe`'s class initializer. As it appears to be set-backed, I don't think the additional creation of a new JeiBrewingRecipe would have any impact on it.

Separate to JEI's code base, there is the potential for mods who have noticed this bug and have implemented their own solution to fix it by registering additional recipes. I'm unaware of any mods that do this, however.

EDIT: There appears to be formatting issues with this commit. When I was publishing to local maven to test in Roots 4, the `build` refused to run unless I executed `:applySpotless`. I didn't see a code style file in order to be able to fix it myself.

### DIFF
--- a/Library/src/main/java/mezz/jei/library/util/BrewingRecipeMakerCommon.java
+++ b/Library/src/main/java/mezz/jei/library/util/BrewingRecipeMakerCommon.java
@@ -38,6 +38,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class BrewingRecipeMakerCommon {
 	private static final Logger LOGGER = LogManager.getLogger();
@@ -122,7 +123,8 @@ public class BrewingRecipeMakerCommon {
 			String inputPathId = PotionSubtypeInterpreter.INSTANCE.getStringName(potionInput);
 
 			for (ItemStack potionReagent : potionReagents) {
-				ItemStack potionOutput = getOutput(potionBrewing, potionInput.copy(), potionReagent);
+				ItemStack potionInputCopy = potionInput.copy();
+				ItemStack potionOutput = getOutput(potionBrewing, potionInputCopy, potionReagent);
 				if (potionOutput.isEmpty()) {
 					continue;
 				}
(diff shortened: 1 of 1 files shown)


mezz:
Thanks for contributing!

Fixes https://github.com/mezz/JustEnoughItems/issues/4035

mezz:
I pushed some changes to make it more "JEI" and confirmed that it's still working well:

mezz:
Non-standard JEI brewing recipes are not impacted by this, the new code is only for those recipes that are created automatically by JEI, so there are no concerns with breaking compatibility.

noobanidus:
Awesome, thank you!

PR #4125, opened 2025-10-25
backport history to 1.20.1

vfyjxf:
backport:#4094 and #3834
Also, allow lookup history to be draggable
image

draggable:
image


mezz:
Thanks!

PR #4133, opened 2025-11-01
Update kk_kz translation

ninsent:
Updated full Kazakh translation


mezz:
Thank you!

Common/src/main/resources/assets/jei/lang/kk_kz.json
```diff
@@ -144,27 +303,41 @@
   "gui.jei.category.smoking": "Ыстау",
   "gui.jei.category.blasting": "Балқыту",
   "gui.jei.category.campfire": "Алауға пісіру",
-  "gui.jei.category.smelting.experience": "%s XP",
-  "gui.jei.category.smelting.time.seconds": "%sс",
-  "gui.jei.category.fuel": "Отын",
+  "gui.jei.category.smelting.experience": "%s тәжірибе",
+  "gui.jei.category.smelting.time.seconds": "%сек",
```
mezz: I think this may be broken, it should have `%s` in it, which will show the number of seconds

mezz (changes requested):
Thank you for contributing!
I think there is one small error, please see my comment

Common/src/main/resources/assets/jei/lang/kk_kz.json
ninsent: Yes, it's definitely a mistake. I accidentally deleted part of the variable. I'll do a commit with the fix soon

ninsent:
> Thank you for contributing! I think there is one small error, please see my comment

I have fixed it

mezz:
Thank you!

Issue #4040, opened 2025-07-12
[Suggestion]: Update/port to new Minecraft versions and new mod loaders


mezz:
### Suggestion

Please don't ask for updates. I don't check the Github issues when I'm working on a major version update, and it's adding extra work for me to go through and close these issues after.

I've been updating this mod to the latest versions out of my own volition for nearly 10 years now; there is absolutely no purpose or benefit to requesting, complaining, pouting, or attempting to manipulate me in the issues to try to convince me it's needed.

Please never ask for mod authors to update their mods, and especially don't act entitled or hostile towards the people working hard on the things you enjoy. Thank you for your understanding.

#### NeoForge:
- [x] https://github.com/neoforged/NeoForge/pull/2021
- [x] Port the JEI code
- [x] Publish to Curseforge and Modrinth

#### Fabric 1.21.10:
- [x] https://github.com/FabricMC/fabric/pull/4841
- [x] https://github.com/FabricMC/fabric/issues/4667
		* ~~WIP Solution: https://github.com/FabricMC/fabric/pull/4878~~
		* New WIP Solution: https://github.com/FabricMC/fabric/pull/4926
- [x] Port the JEI code
- [x] https://github.com/mezz/JustEnoughItems/pull/4134
- [x] Publish to Curseforge and Modrinth

#### Fabric for other older versions than 1.21.10:
* JEI requires Fabric 1.21.10 features to work. Please update to 1.21.10

#### Forge:
- [ ] Gradle improvements (including ForgeGradle in the multiproject currently requires workarounds and slowdowns)
		* https://github.com/MinecraftForge/ForgeGradle/issues/991
		* https://github.com/MinecraftForge/ForgeGradle/issues/992
		* https://github.com/MinecraftForge/ForgeGradle/issues/993
		* Likely more fixes are required but will need multiproject testing on the new ForgeGradle
- [ ] Recipe Sync API from server to client
- [ ] Port the JEI code
- [ ] Publish to Curseforge and Modrinth

mezz:
<img width="493" height="84" alt="Image" src="https://github.com/user-attachments/assets/052d9c49-32ab-4d2d-9015-ca0bb86ed6eb" />
<img width="571" height="65" alt="Image" src="https://github.com/user-attachments/assets/a3562af9-4094-4097-9870-1831b22a49f9" />
[...]

PR #4134, opened 2025-11-02
Hook into fabric recipe sync.

Patbox:
With https://github.com/FabricMC/fabric/pull/4926 being merged and released as Fabric API 0.137.0+1.21.10, JEI can now release updated version on Fabric.

Naturally the mod is needed on server to works (same as neoforge) to enable required recipes.

Changes are pretty simple, since fabric also just syncs Recipe objects.

Solves Fabric side of #4040

### DIFF
--- a/Fabric/src/main/java/mezz/jei/fabric/JustEnoughItems.java
+++ b/Fabric/src/main/java/mezz/jei/fabric/JustEnoughItems.java
@@ -12,11 +12,15 @@ import net.fabricmc.api.ModInitializer;
 import net.fabricmc.fabric.api.recipe.v1.sync.RecipeSynchronization;
 import net.minecraft.core.Registry;
 import net.minecraft.core.registries.BuiltInRegistries;
+import net.minecraft.resources.ResourceKey;
 import net.minecraft.resources.ResourceLocation;
-
-import java.util.Objects;
+import net.minecraft.world.item.crafting.RecipeSerializer;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class JustEnoughItems implements ModInitializer {
+	private static final Logger LOGGER = LogManager.getLogger();
+
 	@Override
 	public void onInitialize() {
 		IServerConfig serverConfig = ServerConfig.getInstance();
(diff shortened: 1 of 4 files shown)


Fabric/src/main/java/mezz/jei/fabric/JustEnoughItems.java
```diff
@@ -24,5 +27,13 @@ public void onInitialize() {
 		var recipeSerializer = new JeiShapedRecipe.Serializer();
 		var registered = Registry.register(BuiltInRegistries.RECIPE_SERIALIZER, resourceLocation, recipeSerializer);
 		RecipeSerializers.register(() -> registered);
+		RecipeSynchronization.synchronizeRecipeSerializer(recipeSerializer);
```
mezz: `JeiShapedRecipe` do not exist on the server, so I think we can remove this

Fabric/src/main/java/mezz/jei/fabric/JustEnoughItems.java
```diff
@@ -24,5 +27,13 @@ public void onInitialize() {
...
+
+		// Run through vanilla recipe serializaers and sync them
+		for (var serializer : BuiltInRegistries.RECIPE_SERIALIZER.keySet()) {
+			if (serializer.getNamespace().equals(ResourceLocation.DEFAULT_NAMESPACE)) {
+				RecipeSynchronization.synchronizeRecipeSerializer(Objects.requireNonNull(BuiltInRegistries.RECIPE_SERIALIZER.getValue(serializer)));
```
mezz: I think it would be better to iterate over entries instead of the `keySet`, so that you don't need `Objects.requireNonNull`

Fabric/src/main/java/mezz/jei/fabric/JustEnoughItemsClient.java
```diff
@@ -23,6 +25,11 @@ public void onInitializeClient() {
 		Translator.setLocaleSupplier(new MinecraftLocaleSupplier());
 		ClientLifecycleHandler clientLifecycleHandler = new ClientLifecycleHandler();
 
+		ClientRecipeSynchronizedEvent.EVENT.register((minecraft, synchronizedRecipes) -> {
+            Internal.setClientSyncedRecipes(RecipeMap.create(synchronizedRecipes.recipes()));
```
mezz:
```suggestion
			Internal.setClientSyncedRecipes(RecipeMap.create(synchronizedRecipes.recipes()));
```
tabs vs spaces

mezz: For the future, it may be nice if the event provided a shared `RecipeMap`, to avoid having multiple mods create it

mezz (changes requested):
Thank you for contributing!

Fabric/src/main/java/mezz/jei/fabric/JustEnoughItems.java
Patbox: Originally did that, but this ended it slightly cleaner (andwith few less lookups I guess). Technically it can't be null anyway, just added that so intelij doesn't warn

Patbox: Okay checked and doing it the other way around still wants Objects#equireNonNull, as Registry#getKey can technically return null.

PR #3803, opened 2024-10-09
Polish translation update

Greg-21:
In this pull request, I updated the Polish translation to include all the changes to the source translation strings made since my last update of the Polish translation 2 years ago in commit dde86189738a28735072ccfa3ee8e425b01c94df. I also took the opportunity to improve some of the existing text strings to ensure the highest possible standards of the translations I maintain.

PR #4142, opened 2025-11-04
Polish translation update

Greg-21:
This pull request includes a *partial* update to the Polish translation, covering only those changes to the source translation strings made since my last update of the Polish translation a year ago in PR #3803 that are completely *unrelated* to configurations, configuration screens, or anything connected to them. I also took the opportunity to improve one of the existing text strings while I was at it.

Getting the configuration part translated properly is a huge task that will take quite a bit of free time, which I just don't have at the moment. I also need to figure out a way to test it in-game, which is another challenge. I'll definitely get down to it eventually, as JEI is too important a mod to leave half-baked, but the timing just isn't right yet. :)


mezz:
Thank you!

PR #4206, opened 2025-12-13
Fix recipe transfer handler for items with custom a subtype interpreter

### LINKED ISSUE #4190: [Bug]: Recipe transfer handler does not work for items with certain subtype interpreters

View a recipe where one of the ingredients has a subtype interpreter that returns different values for `UidContext.Ingredient` and `UidContext.Recipe` whilst accessing the correct workstation for the recipe with all of the required ingredients in your inventory.

### Expected Behavior

Clicking the move items button should transfer the ingredients to the appropriate slots.

### Actual Behavior

The move items button does not work and the missing items tooltip is shown.

Image

### Mod Pack URL (Optional)

_No response_

_No response_

### Extra Notes (Optional)

I did a bit of digging and this seems to be a similar issue to #2102, however the fix was reverted by #2987.

[This comment](https://github.com/mezz/JustEnoughItems/blob/ffe7937d80ca715aceaf797bba22e5f6c906b750/CommonApi/src/main/java/mezz/jei/api/ingredients/subtypes/UidContext.java#L37) in `UidContext.java` still suggests that the recipe transfer handlers are using `UidContext.Recipe`, so I can only assume that this change was unintentional.

I imagine replacing `UidContext.Ingredient` with `UidContext.Recipe` in `RecipeTransferUtil.java` is all that is required.

### latest.log

_No response_

Roelymole:
Very simply, this PR replaces 3 instances of `UidContext.Ingredient` with `UidContext.Recipe` in `RecipeTransferUtil.java` to fix #4190.

### DIFF
--- a/Common/src/main/java/mezz/jei/common/transfer/RecipeTransferUtil.java
+++ b/Common/src/main/java/mezz/jei/common/transfer/RecipeTransferUtil.java
@@ -189,7 +189,7 @@ public final class RecipeTransferUtil {
 
 		for (Map.Entry<Slot, ItemStack> slotTuple : availableItemStacks.entrySet()) {
 			ItemStack slotItemStack = slotTuple.getValue();
-			Object slotItemStackUid = stackhelper.getUidForStack(slotItemStack, UidContext.Ingredient);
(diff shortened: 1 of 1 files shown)


mezz:
Thank you for contributing!

PR #4219, opened 2025-12-27
[1.20.1] Allow to add item bookmark with different nbt

vfyjxf:


mezz:
Thanks!
I will merge this solution for 1.20.1, and use the json serialization solution for later versions.

PR #4220, opened 2025-12-27
[1.21.11] Allow to add item bookmark with different nbt 

vfyjxf:


mezz:
Thanks!
Fixed here:
* https://github.com/mezz/JustEnoughItems/commit/d2cc990e52d5aa21920660ae2d7c3d75aafcdcdd

PR #4221, opened 2025-12-27
[1.21.1] Allow to add item bookmark with different nbt 

vfyjxf:


mezz:
Thanks!
Fixed here:
* https://github.com/mezz/JustEnoughItems/commit/bf2f8f914646fb2b494926b9e691502c818967ba

PR #4224, opened 2025-12-30
Polish translation update

Greg-21:
This is a quick and straightforward update to the Polish translation, focused solely on catching up with Mojang's recent large-scale terminology changes from "Resource Location" to "Identifier" in 1.21.11, as well as in commits 6b615d15ef776abf139339779985a91c59c9c324 and f097882c89e9fbb6ddd1ab0677406fbe4c825df9 on the JEI side.


mezz:
Thank you!

Greg-21:
Thank you for merging my changes from this PR into `1.21.11` as always, but there is also the `26.1` branch for the upcoming Minecraft 26.1 release. May I ask you to cherry-pick (or otherwise port) my changes, perhaps along with the other recently merged translations and those still under review, into the `26.1` branch as well? It would be great to have them included there too so they make it into 26.1, helping avoid unnecessary inconsistencies between releases where there's no real justification for such divergence. However, if that's an unnecessary hassle for you, I can always open a separate PR with the same changes targeting `26.1`. Just let me know, and I'll see what I can do.

Also, while I'm already at it and in case you haven't noticed, a few translations merged last year are still stuck in a limbo called the [`1.21.x` branch](/mezz/JustEnoughItems/commits/1.21.x) (notably Ecuadorian Spanish). It seems that most contributors haven't realized that these new or updated JEI translations aren't actually available in the game, not even their original authors (except for Texaliuz and Xlr11, who later resubmitted their translations in June 2025 and February 2026, respectively). Otherwise, I suspect there would have been more complaints about it in the GitHub Issues section or follow-up PRs by now. I was thinking that perhaps this branch could be removed altogether or repurposed to always point to the latest minor release of the 1.21 line, serving as a stable reference point for contributors to target.

PR #4249, opened 2026-01-20
Update fr_fr.json

ClairDePrince:
Completely reworked French translation with newly added strings


mezz:
Thank you for the update!

About "éléments épinglés", I think this makes sense but it is more like a "pin" in English. 
Since the button icon shows a real bookmark, can you use "signets" or "marque pages"?

bookmark_button_disabled

PR #4263, opened 2026-02-13
Improve Brazilian Portuguese localization for JEI

Xlr11:
Updated Brazilian Portuguese translations for JEI tooltips, key bindings, and configuration options.


mezz:
Thank you!

Issue #4232, opened 2026-01-08
[Suggestion]: Add quick move to IGhostIngredientHandler


Mithi83:
### Motivation and use case:

Being able to move ghost items from JEI to other mods is an incredibly useful feature that I tend to use often in conjunction with AE2, e.g. when setting up filters in storage busses or when partitioning my cells in the workbench. Moving many ghost items via drag & drop however is tedious and error-prone as you constantly have to move the mouse back and forth and thus always have to remember which item you already moved and which is next. It would be much quicker and easier to just shift-click ghost items into place just the way shift-clicking items moves items from player inventory to a chest for example.

I've proposed this idea 2 years ago but sadly never received any feedback (#3321) and given that 26.1 porting just started I'd like to try again to get this feature into JEI, as I'm still convinced that it will improve ease of use.

### Proposed feature:

To make things work, JEI needs to provide an additional function in IGhostIngredientHandler to let mods handle the quick move. I propose adding the following API function

```
        /**
         * Called when a quick-move (e.g. shift-click) is performed to move a ghost item.
         */
        default <I> void quickMove(T gui, ITypedIngredient<I> ingredient) {}
```

with the main design choice rationales being:
* Changing the IGhostIngredientHandler API early during the 26.1 port leaves enough time for mods that need to be ported to 26.1 anyway to adapt to this API change.
* Offering an empty default implementation lowers the bar for porting mods even further. `shouldHighlightTargets()` also offers an empty default implementation.
* The function parameters are analogous to those of `getTargetsTyped(...)` and thus should make it easy for mod authors to implement their handlers.

How to further implement this functionality on the JEI side of things is up for discussion. I'd prefer the default keybinding for the feature to be shift-click to be intuitive for people used to shift-click items into containers, e.g. chests and extend the use of that to ghost items from JEI. AE2 already offers a similar shift-click from the player inventory to ghost item slots.

It should work for the bookmarked items on the left side as well as the items on the right side. What to do with the quick-moved ingredient is up to the other mod.

### Proof-of-concept implementation:

https://github.com/user-attachments/assets/7c08fa71-84ab-4473-ae76-d8b6ca604be0

[...]

PR #4278, opened 2026-03-18
Add quick move support to IGhostIngredientHandler

Mithi83:
This is a more refined implementation of the idea proposed here #4232 to add quick move interactions (i.e. shift-clicking items from JEI into a ghost item handler) to the upcoming 26.1 release to use the window of opportunity a new release offers.

A quick recap of the changes:

- Add a new key binding for quick move defaulting to shift + left-click
- Add a new API function `quickMove(T gui, ITypedIngredient<I> ingredient)` to `IGhostIngredientHandler`
- Provide an empty default implementation of that function in the API to allow users to ignore the addition if they don't want to deal with it
- Add a separate `GhostIngredientQuickMoveManager` and use it in `IngredientGridWithNavigation`


Gui/src/main/java/mezz/jei/gui/ghost/GhostIngredientQuickMoveManager.java
```diff
@@ -0,0 +1,71 @@
...
+			ITypedIngredient<V> ingredient = clicked.getTypedIngredient();
+			List<IGhostIngredientHandler.Target<V>> targets = handler.getTargetsTyped(currentScreen, ingredient, false);
+			if (!targets.isEmpty()) {
+				handlerList.add(handler);
+			}
+		}
+
+		if (handlerList.isEmpty()) {
+			return false;
+		}
```
mezz: I can imagine a case where a handler has 0 targets but still supports quick move, so let's remove the filtering of the  handlerList here.

Gui/src/main/java/mezz/jei/gui/ghost/GhostIngredientQuickMoveManager.java
```diff
@@ -0,0 +1,71 @@
...
+
+		for (IGhostIngredientHandler<T> handler : handlerList) {
+			ITypedIngredient<V> ingredient = clicked.getTypedIngredient();
+			handler.quickMove(currentScreen, ingredient);
```
mezz: `quickMove` should probably return a boolean to indicate if it has handled the move or not. Then this function can return when the first handler has returned `true`.

CommonApi/src/main/java/mezz/jei/api/gui/handlers/IGhostIngredientHandler.java
```diff
@@ -42,6 +42,11 @@ default boolean shouldHighlightTargets() {
 		return true;
 	}
 
+	/**
+	 * Called when a quick-move (e.g. shift-click) is performed to move a ghost item.
+	 */
```
mezz, line 51: note: we'll need to bump the minor version of the mod, and include `@since <version>` in the doc here

I can handle this when the PR is ready to merge

mezz (changes requested):
Thanks for contributing!
This looks like a good feature. I think it could be backported when it's ready, because this can be done without breaking the API.

Gui/src/main/java/mezz/jei/gui/ghost/GhostIngredientQuickMoveManager.java
```diff
@@ -0,0 +1,71 @@
...
+		}
+
+		return true;
+	}
+
+	public <T extends Screen> boolean quickMove(T screen, UserInput input) {
```
mezz, line 38: we'll need to handle the input "simulate" case specially here, to avoid calling the `quickMove` api method twice

mezz (changes requested):

Gui/src/main/java/mezz/jei/gui/ghost/GhostIngredientQuickMoveManager.java
Mithi83: This is based on the code in `GhostIngredientDragManager.handleClickGhostIngredient` where filtering is (probably) used to highlight only matching slots. I'll remove it.

CommonApi/src/main/java/mezz/jei/api/gui/handlers/IGhostIngredientHandler.java
Mithi83: I've left a TODO comment here for visibility.

Gui/src/main/java/mezz/jei/gui/ghost/GhostIngredientQuickMoveManager.java
Mithi83: Agreed and done.

Gui/src/main/java/mezz/jei/gui/ghost/GhostIngredientQuickMoveManager.java
Mithi83: I'll have to research that a bit more before I can propose a fix here.

I've noticed that the call for my API function came twice in my PoC for AE2 but was not sure why. Since I have to filter out duplicates anyway in AE2 to be consistent with quick move from player inventory it would not have been a big deal there but for a general API we should clean that up on the JEI side. I've documented my findings of the PoC in the last bullet point of #4232.

Mithi83: State before any further change: Mouse button pressed called the API and mouse button released called the API, leading to a possibility to receive two API calls with different items if you press the mouse button on item A, move the mouse to item B and release the button there.

I've tried to put a quick `if (input.isSimulate()) {return true;}` in here and it seems to remove the API call when the mouse button is pressed and only triggers an API call on mouse button release. Returning `false` here prevents both events from calling the API, presumably because  `IngredientGridWithNavigation.handleUserInput` only `return Optional.of(this);` if the result of this function here was true. Maybe that should be handled differently as well.

I'm also not sure what would be the appropriate level where this check should be. You could move it up the call stack to `IngredientGridWithNavigation.handleUserInput` with the risk of another potential caller of this function here to miss that check again. You could also push it down further the call stack and in the most extreme case make it part of the API, although I'd prefer not to do it that way for simplicity of the API.

Any thoughts?

mezz: The idea is to "simulate" on click down and execute on click up. 
JEI will make sure that a click is only executed when a user clicks down and lets go on the same element.
This is more similar to how operating systems and web browsers handle clicks. Mojang normally just "clicks" instantly on mouse down, but that leads to misclicks in UI-heavy mods like JEI.

For the logic here, during the "simulate", you just want to return `true` if this element could probably handle the click. This will prevent other elements from trying to handle it.

I think this check should not be done at the API level, it is the responsibility of the element being clicked (or in this case, it could be handled by `GhostIngredientQuickMoveManager` which holds shared logic for those kinds of elements).

Mithi83: Thanks for your explanation, that helped a lot.

Fixing the problem was not as trivial as I thought. Just putting a `return true` at the start of the function would prevent *any* shift-click event from reaching other handlers, in particular preventing a regular shift-click from player inventory in AE2. I therefore had to move it into the part that checked if a ghost item was actually hit by the mouse position. With a bit of boolean magic I was able to come up with this solution. If `input.isSimulate()` is already true the right side of the `||` should never be called and the function will still return `true` in the end.

mezz:
Thanks for contributing!
I fixed up some minor nits and merged this as version 29.1.0

Mithi83:
Great, thanks for the final touches and for merging this PR. I'll post a corresponding PR for AE2 once they port further than snapshot-4.

PR #4286, opened 2026-03-31
1.20.1 async rebuild

Mysticpasta1:
this is at least run twice as fast on multiple report from discord :D and is heading into a major modpack called Integrated MC :D


mezz:
Thanks for the contributions!

Unfortunately I think there are issues with this idea overall.
1. This is combining 3 features into one PR, making it difficult to review. Each feature should be its own PR, with a clear PR description and test results comparing the performance before/after. The performance tests should have detailed enough explanation to be reproducible by anyone who wants to check.
2. JEI Search is very optimized and doesn't need parallel processing to improve it, the data structures involved are more than fast enough for the largest packs in my testing. To justify the added complexity and maintenance, a performance problem should be identified and a strong improvement must be measurable.
3. I am not sure if the tooltip cache is helping or not, or why it is needed. It would need clear performance testing results with and without it.
4. Loading plugins in parallel can lead to serious problems, and some will need to run synchronously on the main thread. To do that without deadlocking or blocking the main thread for a long time, they'll need a way to call a client executor that can schedule bits of work at a time. 
5. Not all JEI data structures are thread-safe, and this PR doesn't seem to identify the main weaknesses or address them.
6. This appears to be mostly written by AI, which means it needs even more careful review than usual. I don't want to take on the extra work to critique AI work. I help review and teach people because they learn and become better programmers. AI can correct issues but it does not learn.
7. I made an async JEI implementation here in 2023 but many mods crashed in testing, and I judged that it might be too complex to support all the issues I'd receive on GitHub: https://github.com/mezz/JustEnoughItems/pull/3219

PR #4296, opened 2026-04-14
add ingredient groups

vfyjxf:
Added a REI-like group feature to JEI, which is currently just working, I can't handle sorting, and I need to incorporate edit mode into this feature and provide a visual editing solution

TODO:

- [ ] Ingame Editor UI
- [ ] dynamic reload
- [ ] Allow to disable some groups 
- [ ] C Tag Translation(common tag has translation,but current design doesn't support it)?

preview:
image

usage:

json def(for modpacker)
config/jei/groups/jei/enchanted_book.json
```json


[
  {
    "version": 1
  },
  {
    "override": true,
    "selectors": [
      {
        "type": "tag",
        "tag": "minecraft:planks"
      },
      {
        "type": "regexp",
        "pattern": "minecraft:book"
      },
      {
        "type": "ingredients",
        "ingredients": [
          {
            "type": "item_stack",
            "ingredient": {
              "id": "minecraft:enchanted_book",
              "components": {
                "minecraft:stored_enchantments": {
                  "minecraft:aqua_affinity": 1
                }
              }
            }
          }
        ]
      }
    ]
  }
]





```


vfyjxf:
I'm not sure what should be used to sort groups and ingredients, so currently groups are adjacent instead of distributed like rei does

vfyjxf:
There are also some features that I'm not sure should be added, such as search prefix for groups and how to let users see which ingredients match without expanding the group

PR #4315, opened 2026-05-22
Allow changing recipe category with shift + scroll

Goldenlion5648:
I made this change for the 1.20.1 branch, but the same patch should be able to cleanly apply to 26.1 as well.

I did the same thing for HadEnoughItems here, and got merged https://github.com/CleanroomMC/HadEnoughItems/pull/194

I'm just spoiled after having this functionality in GTNH's NEI haha


mezz:
Thank you for contributing!

PR #4316, opened 2026-05-23
Update pl_pl.json

Fhilips613:


Greg-21 (changes requested):
Why? Just why?? What have I ever done to you? D:

Most of these changes (aside from the config screen translations) are literally just rewrites of the exact same thing with different word order, synonyms, or alternative verb forms (mostly infinitive → imperative), seemingly just to make it "your way" (or maybe more AI-ish?), without adding any actual value for JEI users. And then all these *pointless* changes have to be backported to every older Minecraft version still supported by JEI. A lot of extra work for absolutely no benefit.

The rest of the changes are, frankly, clear regressions compared to what's currently there and what I originally contributed. Most likely because the translation was done without any real understanding of the context in which these strings actually appear in-game, so everything was essentially translated by gut feeling. Just take a look at `jei.tooltip.liquid.flowing` or `jei.tooltip.recipe.by`, and that's only within the first 20 lines! Not to mention something as basic as mixing up smelting and blasting...

The Polish JEI translation currently in use is something I've been gradually refining, little by little, over many years (and compared to most translators, I'm probably one of the very few who actually kept actively maintaining it; just look at the [contributor stats](../graphs/contributors), I'm the No. 2 JEI contributor), and I know for a fact that it's very well-made and carefully thought out.

So please, I'm begging you, don't fix what isn't broken! There are so many projects out there with no Polish translation at all that could genuinely benefit from your help and contributions, instead of wasting that effort "improving" what is arguably one of the best Polish Minecraft mod translations out there (right alongside the Alex's Caves Polish translation by @Shadow75857, which is honestly excellent and I genuinely can't find a single issue with).

Fhilips613:
I'm so so sorry i saw that there mod's translation is outdated. Like en_us had like 355 or smth and pl_pl had like 200 so I wanted to translate it. If it isn't good than sorry I tried 💔

Greg-21:
I'm sorry if I sounded harsher than I should have, but this isn't the first time recently when you've tried to "improve" my translations and ended up suggesting something that, for the most part, isn't really an improvement. And I'm just slowly getting tired of it.

Of course, nobody is infallible, myself included, so I'm always open to actual constructive discussions on what could be improved, why, and how. But making pointless changes and randomly replacing tons of perfectly fine sentences with basically the same thing just rewritten using different word order, synonyms, or alternative verb forms is definitely not the way to go here.

You said that you noticed that some strings present in the U.S. English translation were missing from the Polish one and wanted to help with that, which I appreciate. What I don't understand, though, is why you also went ahead and touched a bunch of completely unrelated lines. From my perspective, it honestly kind of looks like you just dumped the whole file into some LLM, took the output, tweaked it a bit to your liking, and committed it. You could've just translated the newly added strings, left the existing ones alone, and called it a day. Sure, figuring out what's new and what's old takes some extra effort, but other translators manage to do exactly that just fine, so I'm sure you could handle it too. Don't look at me, though, as I happen to have a custom-made tool that I wrote, among other things, for this exact purpose, which automates the entire process and makes it a pleasure ;)

As for calling the Polish JEI translation "outdated", well, I can see where you're coming from. There are indeed quite a few missing strings related to configurations, configuration screens, and anything connected to them. Thankfully (at least for me), it's also one of those areas that, while fairly large, is rarely used by most players, since the vast majority just stick to the default settings regardless of what they happen to be.

Also, the exposure surface for those strings was significantly reduced after Forge 1.13 removed the graphical configuration GUI in 2019, and even though NeoForge 1.21 finally brought it back in 2024 after many, many years of absence, Mezz has not decided to make use of it in JEI so far, as it is not extensible enough to hook it up to the custom configuration system used by JEI (if you'd like to read more on that, feel free to check out PR four thousand ninety-five here on GitHub). So nowadays, most of those English-only strings mainly appear only inside the text config files located in `~/.minecraft/config/jei/`, specifically in the comments contained within those files.

I'm fully aware that this is still an issue (perhaps not a major one, but still), but like I already mentioned in one of the earlier PRs (I'll just quote my past self here):

> Getting the configuration part translated properly is a huge task that will take quite a bit of free time, which I just don't have at the moment. I also need to figure out a way to test it in-game, which is another challenge. I'll definitely get down to it eventually, as JEI is too important a mod to leave half-baked, but the timing just isn't right yet. :)

PR #4321, opened 2026-05-27
[26.1] Polish translation update

Greg-21:
In this pull request, I updated the Polish translation to include the new text string introduced in commit 0639eebf3085b480e4a71ebc9fcbdba87de502a1, as well as the "Resource Location" → "Identifier" terminology changes from PR #4224, which were only merged for 1.21.11 and never made it into 26.1 or 26.2, leaving these two recent versions slightly behind in this regard.


mezz:
Thank you!

Greg-21:
Thank you for this fantastic mod, as always! :)

PR #4324, opened 2026-06-02
[26.1] Port Amecs support to fabric 26.1

bluegreensea:
Some other changes:
Configured support update (but Configured still required update)
In FabricKeyMapping to using key instead of realKey (Improve compat and after Vanilla 1.21.9 solved should not required)

known bugs:
binded key combination will blocked in game


mezz:
Thank you for contributing!
I fixed up some issues and resolved the problem where bound keys would prevent their use in-game.

PR #4333, opened 2026-06-10
Fix 9th slot of the hotbar can not be recognized for Smithing Table recipe transfer

qwqmb:
Fixed an issue where the 9th slot of the hotbar could not be recognized when using recipe transfer for the Smithing Table. Fixed and compiled tested in 1.20.1. Other versions also have this issue but have not been tested.


mezz:
Thank you for contributing!

Issue #4245, opened 2026-01-18
[Bug]: Mod id formatting on fluids breaks if both JEI and Jade are installed


Maxi3773:
### Steps to Reproduce the Bug

1. Install JEI and Jade
2. Hover over any fluid in JEI

| Expected (only JEI) | Actual (JEI + Jade) |
| ----- | ----- |
|<img width="400" height="200" alt="Image" src="https://github.com/user-attachments/assets/8b4751f1-c900-4a33-b893-962c6486ed3a" />| <img width="400" height="200" alt="Image" src="https://github.com/user-attachments/assets/492e5d72-0a9a-4e25-b006-060f3ec98c45" />|

### Extra Notes

**According to the developer of Jade, this bug is caused by JEI, not Jade.** (See Snownee/Jade#545)

Tested on NeoForge 21.1.218 and the latest versions of the mods for 1.21.1.

### latest.log

https://gist.github.com/Maxi3773/1de421cf4e7a0f967d60fe0104017920

Maxi3773:
Since the bot warned about this issue being closed, I would kindly like to ask whether a fix is planned and thus I should keep the issue open or not?

mezz:
Thanks for the report!
This issue has been fixed on 1.21.1 and 26.2

PR #4343, opened 2026-06-24
[1.21.1] Fix mod-name formatting detection

0196071:
Fix mod-name formatting detection when another mod adds a styled Component tooltip line.

Previously, JEI detected existing mod-name tooltip formatting using Component#getString(). This preserves legacy formatting codes in the text, but does not preserve styles applied directly to the Component. As a result, when another mod added a styled mod-name line without legacy formatting codes, JEI detected the override as plain %MODNAME%. Item tooltips still looked correct because the other mod added the item mod name, but JEI-rendered non-item ingredient tooltips, such as fluids, used the plain override.

This change recovers legacy formatting from the Component style when the detected format would otherwise be plain %MODNAME%.

Fixes https://github.com/mezz/JustEnoughItems/issues/4245 , https://github.com/Snownee/Jade/issues/545

### DIFF
--- a/Library/build.gradle.kts
+++ b/Library/build.gradle.kts
@@ -33,6 +33,7 @@ dependencyProjects.forEach {
 
 neoForge {
     neoFormVersion = "$minecraftVersion-$neoformTimestamp"
+    addModdingDependenciesTo(sourceSets.test.get())
 }
 
 sourceSets {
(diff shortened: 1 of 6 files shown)


mezz:
Thanks for contributing!
I added a lot of tests and fixed some weird cases, hopefully this'll solve any future issues too.

PR #4350, opened 2026-07-02
Client-side recipe fallback + fix empty-recipe GUI crash

CervantesHallo:
# PR: Client-side recipe fallback + empty-recipe crash fix (26.1)

Branch: `clientside-recipe-fallback` (2 commits, off `26.1`)

## Motivation
Since Minecraft 1.21.2 recipes are server-authoritative and reach JEI through
Fabric's `ClientRecipeSynchronizedEvent`. On servers that never send it — most
notably older-version servers reached through **ViaVersion** — JEI receives no
recipes. Two problems follow:

1. **Crash:** opening a container makes JEI start with zero recipe categories.
   Building the recipe GUI indexes into the empty category list
   (`IngredientLookupState.getFocusedRecipes` → `recipeCategories.get(0)`),
   throwing `ArrayIndexOutOfBoundsException` during GUI startup — which aborts
   the whole runtime registration, so even the item-list overlay disappears.
2. **No recipes:** even without the crash, there is nothing to browse.

## Changes
**Commit 1 — crash fix (safe, standalone):**
- `RecipeGuiLogic.hasMultiplePages()` and `IngredientLookupState.recipeCount()`
  return early when the recipe-category list is empty.

**Commit 2 — client-side recipe fallback (opt-in behavior, Fabric only):**
- New `ClientRecipeLoader`: when the sync event did not happen, load the client's
  own built-in vanilla recipe files (`data/*/recipe/*.json`) via a `SERVER_DATA`
  `MultiPackResourceManager`, deserialize with the connected server's registries
  (`connection.registryAccess().createSerializationContext(JsonOps.INSTANCE)`),
  and feed them to JEI as client-synced recipes.
- `ClientLifecycleHandler`: in the existing container-open fallback, load
  client-side recipes if none are present, then re-initialize the already-open
  screen so JEI's per-screen input handlers (registered on `BEFORE_INIT`) attach.

## Notes / open questions for maintainers
- Commit 1 is a pure defensive fix and is useful regardless of the rest.
- Commit 2 intentionally re-introduces client-side recipe reading as a *fallback
  only* (server-synced recipes still take precedence via `AFTER_RECIPE_SYNC`).
  If reading client recipes on non-syncing servers is undesirable by design, it
  could be gated behind a config option, or Commit 1 could be merged alone.
- Only vanilla client recipes are loaded currently; modded data packs are not
  included in the fallback resource manager.

### DIFF
--- a/Fabric/src/main/java/mezz/jei/fabric/startup/ClientLifecycleHandler.java
+++ b/Fabric/src/main/java/mezz/jei/fabric/startup/ClientLifecycleHandler.java
@@ -12,6 +12,7 @@ import mezz.jei.library.startup.StartData;
 import net.fabricmc.fabric.api.client.keymapping.v1.KeyMappingHelper;
 import net.fabricmc.fabric.api.client.screen.v1.ScreenEvents;
 import net.minecraft.client.Minecraft;
+import net.minecraft.client.gui.screens.Screen;
 import net.minecraft.client.gui.screens.inventory.AbstractContainerScreen;
 import net.minecraft.server.packs.resources.ResourceManagerReloadListener;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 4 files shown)


mezz:
Thank you for contributing!

Sorry, I was already working on my own version of this and merged it here: https://github.com/mezz/JustEnoughItems/commit/274fd56a841d3d4d3fea703eb4b93e4eb4554437
It took me a while because I wanted to write a lot of automated tests for it.

PR #4356, opened 2026-07-06
[Feature] Add option to disable toast notification reflow

### LINKED ISSUE #4186: [Suggestion]: Option to not reflow from toast notifications
### Suggestion

Toasts make the entire index shift over to the left when they popup, rather than down, which can get a bit annoying. Would be nice to be able to disable this behavior entirely and get the old overlap instead

Oleksii-Melashchenko:
**Summary:**
Added a new configuration option `toastReflowEnabled` to allow players to disable the automatic reflow of the JEI interface when toast notifications appear.

**Reasoning:**
By default, JEI shifts its item index to avoid overlapping with incoming toast notifications. When multiple toasts appear simultaneously this shifting can be visually disruptive. This change gives users the option to disable this behavior entirely.

**Important Note:**
When `toastReflowEnabled` is set to `false`, the JEI index will remain stationary, restoring the classic overlapping behavior requested by users. The toast notifications will overlap with the JEI interface, giving players full control over their preferred UI layout.
Closes #4186

### DIFF
--- a/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
+++ b/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
@@ -22,6 +22,7 @@ public final class ClientConfig implements IClientConfig {
 	// appearance
 	private final Supplier<Boolean> centerSearchBarEnabled;
 	private final Supplier<Integer> maxRecipeGuiHeight;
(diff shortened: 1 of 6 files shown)


mezz:
Thanks for contributing!
It looks like a good config to have.

I updated the PR with some minor fixes.

PR #4385, opened 2026-07-25
Fix ingredient grid anchor reset on overlay close

vfyjxf:
see this gif:
202607206235810_small

Based on the commit history, I suspect this [commit](https://github.com/mezz/JustEnoughItems/commit/7d1237fffe3e64c90ac60efeef2b7469b236bb11) caused the issue.


mezz:
Thanks for contributing!
I added some tests to catch this problem and your fix works. Looks good to me.

PR #4386, opened 2026-07-25
Fix: call `super.extractRenderState`

cassiancc:
Unlike the 26.2 version, JEI's Recipe GUI for 26.1 does not call `super.extractRenderState`, which led to hard to diagnose compatibility issues with mods that expect vanilla `Screen` methods to be called, like the [method that renders vanilla renderables](https://mcsrc.dev/1/26.1.2/net/minecraft/client/gui/screens/Screen#L126-128).


mezz:
Thanks for contributing!
I just backported this in a big batch of backporting though, sorry didn't see this in time.

PR #4387, opened 2026-07-26
Fix #4381 keep page position on resize

### LINKED ISSUE #4381: [Bug]:[1.12.2] JEI list jumps to first page after resize when GUI extends beyond normal size

Open a larger GUI, such as the Diamond Chest from Iron Chests, or the Large Backpack from Useful Backpacks.

### Expected Behavior

JEI should keep its original interface.

### Actual Behavior

Actually, JEI resets to the first page.

### Mod Pack URL (Optional)

_No response_

_No response_

### Extra Notes (Optional)

This bug occurs in jei_1.12.2-4.16.5.1015 and still exists in jei_1.12.2-4.16.5.1016. Please see the video below for the actual effect.

https://github.com/user-attachments/assets/0ee2aae5-33d0-4824-9341-d187a887aa6e

### latest.log

_No response_

vfyjxf:
Port the design from 26.2, along with the bug fixes from #4385

### DIFF
--- a/src/main/java/mezz/jei/gui/overlay/IngredientGridWithNavigation.java
+++ b/src/main/java/mezz/jei/gui/overlay/IngredientGridWithNavigation.java
@@ -36,6 +36,8 @@ public class IngredientGridWithNavigation implements IShowsRecipeFocuses, IMouse
 	private static final int NAVIGATION_HEIGHT = 20;
 
 	private int firstItemIndex = 0;
+	@Nullable
+	private IIngredientListElement pageAnchorElement;
 	private final IPaged pageDelegate;
 	private final PageNavigation navigation;
 	private final GuiScreenHelper guiScreenHelper;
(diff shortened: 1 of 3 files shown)


src/main/java/mezz/jei/render/IngredientListBatchRenderer.java
```diff
@@ -46,6 +46,15 @@ public int size() {
...
 		slots.add(ingredientListSlot);
+		if (ingredientListSlot.isBlocked()) {
+			blocked++;
+		}
+	}
+
+	public void add(List<IngredientListSlot> ingredientRow) {
+		for (IngredientListSlot ingredientListSlot : ingredientRow) {
+			add(ingredientListSlot);
+		}
```
vfyjxf: Although it is not necessary here, it is mainly added to maintain compatibility with my mod JEI Utilities.

PR #4389, opened 2026-07-26
Add keybinding for transfer in recipes gui

vfyjxf:
Add keybinding for transfer in recipes gui，and added a config entry to control whether to display recipe transfer errors when mouse over recipe layout.
display:
图片


mezz:
Thanks for contributing!
I think the keybinding text should only show up when the recipe transfer is possible.
When showing "Missing Items", the keybinding instructions seem to contradict it, which could be confusing.

vfyjxf:
> Thanks for contributing! I think the keybinding text should only show up when the recipe transfer is possible. When showing "Missing Items", the keybinding instructions seem to contradict it, which could be confusing.

I added these tooltips mainly as hints; otherwise, users might have a hard time noticing them. Maybe it would be better to put them on the transfer button?

Common/src/main/resources/assets/jei/lang/en_us.json
```diff
@@ -31,6 +31,8 @@
   "jei.tooltip.bookmarks.tooltips.usage": "[Press \"%s\" to show details]",
   "jei.tooltip.bookmarks.tooltips.transfer.usage": "[Press \"%s\" to craft one]",
   "jei.tooltip.bookmarks.tooltips.transfer.max.usage": "[Press \"%s\" to craft many]",
+  "jei.tooltip.recipesGui.tooltips.transfer.usage": "[Press \"%s\" to craft one]",
+  "jei.tooltip.recipesGui.tooltips.transfer.max.usage": "[Press \"%s\" to craft many]",
```
Greg-21: First, I would suggest avoiding duplicating identical translation strings where possible.

In my experience, this kind of 1:1 duplication usually ends up with the two copies getting out of sync sooner or later. Someone updates one string but forgets to update the other, or translates one of them one day and comes back to the other much later, unintentionally ending up with slightly different wording between the two, or simply doesn't notice that they were identical in U.S. English and should probably stay that way in localized translations as well.

Admittedly, that's fairly unlikely to happen right now since the entries are right next to each other, but there's no guarantee they will always stay that way. To be on the safe side, I'd rather avoid introducing that maintenance burden in the first place by simply reusing a single translation key and its associated string for both use cases instead.

Something along these lines:
```suggestion
  "jei.tooltip.transfer.usage": "[Press \"%s\" to craft one]",
  "jei.tooltip.transfer.max.usage": "[Press \"%s\" to craft many]",
```

Common/src/main/resources/assets/jei/lang/en_us.json
```diff
@@ -97,6 +99,8 @@
   "key.jei.showUses2": "Show Uses",
   "key.jei.transferRecipeBookmark": "Craft Bookmarked Recipe (One)",
   "key.jei.maxTransferRecipeBookmark": "Craft Bookmarked Recipe (Many)",
+  "key.jei.transferRecipeGui": "Craft Recipe (Recipes GUI)",
+  "key.jei.maxTransferRecipeGui": "Craft Recipe (Many) (Recipes GUI)",
```
Greg-21: I also have one small suggestion regarding the new keybinding names after building the current WIP version from this branch and trying it in-game to see how everything looked in practice.

Since you mentioned that these tooltips are mainly intended as hints to help users discover the feature, I don't think the corresponding keybinding names need to be quite that explicit. For example, "Craft Recipe (Many) (Recipes GUI)" reads a little awkwardly in English because of the nested parentheses.

Since they're already grouped under the "JEI (Recipes)" category in the Key Binds screen, that alone already provides enough context for users to understand where those keybindings apply, I think.

With all that in mind, I'd probably go with something like this:
```suggestion
  "key.jei.transferRecipeGui": "Craft Recipe (One)",
  "key.jei.maxTransferRecipeGui": "Craft Recipe (Many)",
```

Or alternatively:
```suggestion
  "key.jei.transferRecipeGui": "Craft Selected Recipe (One)",
  "key.jei.maxTransferRecipeGui": "Craft Selected Recipe (Many)",
```

Greg-21 (changes requested):
If I may add my two cents, I have a couple of small suggestions for your consideration.

Common/src/main/resources/assets/jei/lang/en_us.json
vfyjxf: I am not a native English speaker, which one do you think is better 🤔

vfyjxf: Adding "selected" might be more detailed, but this button is usually displayed in the Recipes GUI via a tooltip, so it may not need to be emphasized.

PR #4391, opened 2026-07-26
Utilize `getCreatorNamespace` on Fabric

cassiancc:
Fabric has had a method equivalent to NeoForge's `getCreatorModId` since 1.21.8, and I backported it to 1.21.1, 1.21.4, and 1.21.5 when it was initially released.
https://github.com/FabricMC/fabric-api/pull/4746


mezz:
Thanks for contributing! I'll backport this to older supported versions.

PR #4423, opened 2026-08-14
Fix IME preedit rendering in the ingredient search field

### LINKED ISSUE #4421: [Bug]: IME preedit text is not rendered in the JEI search field

1. Enable an IME that uses composition text, such as Korean, Japanese, or Chinese.
2. Open an inventory with the JEI ingredient list visible.
3. Focus the JEI ingredient search field.
4. Start composing text without committing it.

This can also be reproduced in the creative inventory while the creative search tab is selected.

### Expected Behavior

The IME preedit text should be rendered at the cursor in the JEI ingredient search field.

In the creative inventory, the preedit overlay should remain in the JEI search field when that field has focus.

### Actual Behavior

The JEI search field does not receive the screen's preedit events, so composition text is not rendered there.

In the creative inventory, Minecraft forwards preedit events directly to its own search box, causing the preedit overlay to appear in the creative inventory search field instead of the focused JEI search field.

### Mod Pack URL (Optional)

_No response_

_No response_

### Extra Notes (Optional)

`GuiTextFieldFilter` extends Minecraft's `EditBox`, but it is rendered and managed outside the normal `Screen` listener hierarchy.

Additionally, `CreativeModeInventoryScreen#preeditUpdated` bypasses the currently focused screen listener and forwards the event directly to its own search box.

This affects both Fabric and NeoForge.

### latest.log

_No response_

shihyeon:
Fixes #4421 

## Summary

- Register the JEI ingredient search field as the current screen focus target while it is focused.
- Notify Minecraft when the JEI search field gains or loses text input focus.
- Route creative inventory preedit events to the JEI search field when it owns the screen focus.
- Preserve the vanilla creative search behavior when JEI does not have focus.
- Add the creative inventory handling for both Fabric and NeoForge.

## Cause

The JEI search field extends Minecraft's `EditBox`, but it is managed outside the screen's normal listener hierarchy. As a result, IME preedit events sent through the active screen do not normally reach it.

`CreativeModeInventoryScreen` also overrides `preeditUpdated` and sends the event directly to its own search box, bypassing the focused screen listener.

### DIFF
--- a/Fabric/src/main/java/mezz/jei/fabric/mixin/CreativeModeInventoryScreenMixin.java
+++ b/Fabric/src/main/java/mezz/jei/fabric/mixin/CreativeModeInventoryScreenMixin.java
@@ -1,0 +1,26 @@
+package mezz.jei.fabric.mixin;
+
+import mezz.jei.gui.input.GuiTextFieldFilter;
+import net.minecraft.client.gui.components.events.GuiEventListener;
+import net.minecraft.client.gui.screens.inventory.CreativeModeInventoryScreen;
+import net.minecraft.client.input.PreeditEvent;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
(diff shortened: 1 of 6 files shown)


mezz:
This fix has been backported to 26.1. now

PR #4425, opened 2026-08-15
Fix #4421 Fix IME input in search field

### LINKED ISSUE #4421: [Bug]: IME preedit text is not rendered in the JEI search field

mezz:
Fixes:
* https://github.com/mezz/JustEnoughItems/issues/4421

This PR is similar to #4422, but it works with all modded guis (not just creative menu) and adds tests.
* https://github.com/mezz/JustEnoughItems/pull/4422

I will probably follow-up on this to create a NeoForge hook for this instead of using a mixin there.

Requesting review from @shihyeon ([koreanify](https://github.com/erudites-dev/koreanify)) and @vfyjxf ([JustEnoughCharacters](https://github.com/Towdium/JustEnoughCharacters))


vfyjxf (approved):
LGTM

shihyeon:
looks good. ill give it a try

mezz:
Thanks!
@shihyeon should I merge this so you can test, or are you testing it in dev environment now?

shihyeon:
I tested it, and focusing the input field works well. The previous conflict between vanilla mc and IME on/off switching has also been resolved.

shihyeon:
Hold on, there seems to be a conflict with Koreanify. However, this looks like an issue on Koreanify’s side, so I’ll look into it and get back to you.

shihyeon:
The issue with Koreanify has been resolved on my end. However, the issue mentioned below is still unresolved, and I’m currently trying to find the cause.
> The previous conflict between vanilla mc and IME on/off switching has also been resolved.

shihyeon:
I’ve submitted a fix in #4426

mezz:
I opened the follow-up PR to NeoForge here:
* https://github.com/neoforged/NeoForge/pull/3410

shihyeon:
Looks like NeoForge recommends using events. I mainly develop for Fabric and only maintain the NeoForge version, so I wasn’t really familiar with that. lol

mezz:
Using events is nice, it will allow mods to interact in a more complicated way (which is a goal of NeoForge).
Multiple mods (koreanify, JEI, and others) will have a better chance of working together correctly if they can all use the same well-defined event. When NeoForge is ported to new Minecraft versions, the event will also be ported, and hopefully that means less coordination is required between mods on every update.

PR #4428, opened 2026-08-15
Add config toggle to disable recipe sync warning messages

### LINKED ISSUE #4397: [Suggestion]: Add config toggle to disable Recipe Sync Warning Messages
### Suggestion

This Paper (Velocity) server does not provide recipes to JEI. JEI is showing default recipes from your client, which may be different on the server.

This message pops up frequently when switching servers via Velocity.

MakerYuichi:
Fixes #4397

Adds a new `recipeSyncWarningEnabled` boolean under the `advanced` config category, defaulting to `true` so existing behavior is unchanged unless a player opts out.

When disabled, all four recipe-sync chat warnings in `JeiStarter.verifyClientRecipes()` are suppressed in-game (sync error, JEI missing on server, vanilla server, sync unavailable) the issue specifically describes the "vanilla server" case (Paper/Velocity servers that don't sync recipes to JEI), but all four messages share the same root cause, so one toggle covers the category rather than just the one variant.

Each message is still logged to console/log file regardless of the toggle, so the information isn't lost — only the in-game chat popup is suppressed.

**Testing:** `TestClientConfig` (the test double for `IClientConfig`) updated to implement the new interface method. Full test suite passes (`./gradlew test`). Manually verified the config value round-trips through JEI's config file.

### DIFF
--- a/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
+++ b/Common/src/main/java/mezz/jei/common/config/ClientConfig.java
@@ -44,6 +44,7 @@ public final class ClientConfig implements IClientConfig {
 	// advanced
 	private final ConfigValue<Boolean> lowMemorySlowSearchEnabled;
 	private final ConfigValue<Boolean> catchRenderErrorsEnabled;
+	private final ConfigValue<Boolean> recipeSyncWarningEnabled;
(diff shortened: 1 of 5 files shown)


mezz:
Thanks for contributing!

PR #4441, opened 2026-08-22
Shrink the retained GUI item atlas after JEI screens close

Watchedfungus88:
On Fabric 26.2 with the Vulkan renderer, opening an inventory with JEI can lower FPS even after the screen is closed. Reopening it makes the slowdown progressively worse until it plateaus, especially when enchanted items are visible in the HUD.

JEI ingredient rendering can grow the cached `GuiItemAtlas` substantially. Minecraft keeps that larger atlas afterward, while animated foil entries are discarded and drawn back into it every frame.

This adds a small Fabric mixin that closes the cached atlas when the current frame needs a smaller power-of-two texture. The existing vanilla code then recreates it at the right size.

The 26.2 Fabric jar builds cleanly.

### DIFF
--- a/Fabric/src/main/java/mezz/jei/fabric/mixin/GuiRendererMixin.java
+++ b/Fabric/src/main/java/mezz/jei/fabric/mixin/GuiRendererMixin.java
@@ -1,0 +1,34 @@
+package mezz.jei.fabric.mixin;
+
+import net.minecraft.client.gui.render.GuiItemAtlas;
+import net.minecraft.client.gui.render.GuiRenderer;
+import org.jspecify.annotations.Nullable;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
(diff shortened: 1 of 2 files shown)


mezz:
Thanks for contributing!

Unfortunately I have some reservations about this change and cannot accept it. 

* the AI generated description doesn't give me high confidence.
* this workaround is global, not limited to JEI. In general I must avoid impacting the game in unexpected ways with JEI because it is such a common mod to use.
* the workaround is Fabric-only, and JEI is intended for all mod loaders.

Since this is a global change that can impact other mods, and the issue you identified can be trigged by other scenarios (like scrolling through the creative menu or a large crafting storage system), I think it would be better suited to a "performance" mod instead of JEI.

Watchedfungus88:
Thanks for reviewing. I used AI to help write the description because I only speak Russian. The issue and FPS measurements came from my own testing, so the AI-assisted wording should not affect whether the reported behavior is real.

PR #4497, opened 2026-09-19
[1.12.2] Expose recipe id tooltip callback

jbredwards:
Adds ~~two new methods to `IGuiIngredientGroup` (`addRecipeIdTooltipCallback(ResourceLocation)` and `createRecipeIdTooltipCallback(ResourceLocation)`)~~ a new `IRecipeLayout.setRecipeId(ResourceLocation)` method, to make it easier to add recipe id and "recipe by" text to recipe category output ingredient tooltips. This will also allow any future changes made to the tooltip callback to carry over to modded recipe categories!

This PR also fixes a bug where the crafting recipe "recipe by" tooltip callback checked the item's registry mod id instead of its creator mod id, which caused it to sometimes render when it shouldn't (like for forge's universal bucket item if the contained fluid had the same mod id as the recipe).


mezz:
Thanks for contributing!
Can you please explain why you need to override this? Ideally please point me to your mod's code so I can understand better.

jbredwards:
> Can you please explain why you need to override this?

The main goal of this PR is to make it easier for custom recipe categories with recipe ids to be consistent with the crafting recipe category, without needing to copy/paste its tooltip callback code directly (which can change, as is the case for a JEI fork called [Had Enough Items](https://github.com/CleanroomMC/HadEnoughItems), which I'll create a similar PR for if this one is merged).

> Ideally please point me to your mod's code so I can understand better.

https://github.com/jbredwards/Smithing-Table/blob/2641a7003133fb42dde97d048b4d92c1d91fe2c8/src/main/java/git/jbredwards/smithing_table/mod/common/compat/justenoughitems/SmithingJEIRecipeCategory.java#L97-L110

Thank you for still supporting 1.12! ❤️

src/api/java/mezz/jei/api/gui/IGuiIngredientGroup.java
```diff
@@ -59,6 +61,35 @@ public interface IGuiIngredientGroup<T> {
...
+	 * the tooltip for these ingredients that aren't marked as "input", as seen
+	 * in the crafting recipe category.
+	 * <p>
+	 * Should be used by recipe categories whose recipes have an id.
+	 *
+	 * @param recipeId registry id of the recipe
+	 * @see IGuiIngredientGroup#addTooltipCallback(ITooltipCallback)
+	 * @since JEI 4.16.6
+	 */
+	ITooltipCallback<T> createRecipeIdTooltipCallback(@Nullable ResourceLocation recipeId);
```
mezz: Is this method needed?

mezz: Also, on 1.12.2 there are unfortunately mods that implement `IGuiIngredientGroup`, so we'll need all new api methods to have defaults

mezz (changes requested):
This seems like a good improvement but needs some small changes for the API

jbredwards:
I think I've come up with a better approach! Now it only adds one new `IRecipeLayout.setRecipeId` method to the api, and also adds the new tooltip after any existing tooltip callback instead of replacing it

src/main/java/mezz/jei/gui/recipes/RecipeLayout.java
```diff
@@ -334,4 +336,22 @@ public int getPosX() {
...
+		this.recipeId = recipeId;
+	}
+
+	private void addRecipeIdTooltip() {
+		if (recipeId != null) {
+			ResourceLocation recipeCategoryId = new ResourceLocation(recipeCategory.getUid());
+			boolean recipeCategoryIdDifferent = !recipeId.getNamespace().equals(recipeCategoryId.getNamespace());
+
+			for (GuiIngredientGroup guiIngredientGroup : guiIngredientGroups.values()) {
+				IIngredientHelper ingredientHelper = guiItemStackGroup.getIngredientHelper();
```
mezz: this always retrieves `guiItemStackGroup.getIngredientHelper()`. For a fluid or custom output, `RecipeIdTooltipCallback` passes that ingredient to the item-stack helper, causing a `ClassCastException` when “recipe by” is enabled.

src/main/java/mezz/jei/gui/recipes/RecipeLayout.java
```diff
@@ -334,4 +336,22 @@ public int getPosX() {
...
 	}
+
+	@Override
+	public void setRecipeId(@Nullable ResourceLocation recipeId) {
+		this.recipeId = recipeId;
+	}
+
+	private void addRecipeIdTooltip() {
+		if (recipeId != null) {
+			ResourceLocation recipeCategoryId = new ResourceLocation(recipeCategory.getUid());
```
mezz: The 1.12 API defines category IDs as strings and its built-ins use forms such as `minecraft.crafting`. That means a modded ID like `example.crusher` will parse here with the default `minecraft` namespace, which isn't right. The same-mod comparison becomes incorrect and can add unwanted “recipe by” text.

src/main/java/mezz/jei/gui/recipes/RecipeLayout.java
```diff
@@ -28,10 +30,7 @@
 import mezz.jei.gui.Focus;
 import mezz.jei.gui.TooltipRenderer;
 import mezz.jei.gui.elements.DrawableNineSliceTexture;
-import mezz.jei.gui.ingredients.GuiFluidStackGroup;
-import mezz.jei.gui.ingredients.GuiIngredient;
-import mezz.jei.gui.ingredients.GuiIngredientGroup;
-import mezz.jei.gui.ingredients.GuiItemStackGroup;
+import mezz.jei.gui.ingredients.*;
```
mezz: please avoid * imports

src/main/java/mezz/jei/gui/ingredients/GuiIngredientGroup.java
```diff
@@ -170,4 +170,23 @@ public void setOverrideDisplayFocus(@Nullable IFocus<T> focus) {
...
+	 * @since JEI 4.16.6
+	 */
+	public IIngredientHelper<T> getIngredientHelper() {
+		return ingredientHelper;
+	}
+
+	/**
+	 * Adds the tooltip callback after any existing tooltip callback
+	 * @since JEI 4.16.6
+	 */
```
mezz: this class is internal (not in the api), so it doesn't need javadocs with `@since `

src/api/java/mezz/jei/api/gui/IRecipeLayout.java
```diff
@@ -70,6 +72,18 @@ public interface IRecipeLayout {
 	 */
 	void setShapeless();
 
+	/**
+	 * Adds recipe id and "recipe by" text to all output ingredients of the recipe, as seen in crafting table recipes.
+	 * <p>
+	 * Should be used by recipe categories whose recipes have an id.
+	 *
+	 * @param recipeId registry id of the recipe
+	 * @since JEI 4.16.6
```
mezz, line 82: you'll need to bump the version to 4.16.6 in `gradle.properties`

mezz (changes requested):

mezz (approved):
Looks good, thanks!

